PR #1, opened 2020-09-18
Remove hardcoded instances of user role

jdgarrett:
Also allow users with multiple roles to switch orgs.

### DIFF
--- a/server/api/user/user.model.ts
+++ b/server/api/user/user.model.ts
@@ -3,6 +3,7 @@ import {
 } from 'typeorm';
 import { Role } from '../role/role.model';
 
+
 @Entity()
 export class User extends BaseEntity {
 
@@ -25,15 +26,17 @@ export class User extends BaseEntity {
   })
   roles: Role[];
 
-  @Column({
-    nullable: true,
-  })
+  @Column()
   first_name: string;
 
-  @Column({
-    nullable: true,
-  })
+  @Column()
   last_name: string;
+
+  @Column()
+  phone: string;
+
+  @Column()
+  email: string;
 
   @Column({
     default: true,
@@ -45,24 +48,28 @@ export class User extends BaseEntity {
   })
   root_admin: boolean;
 
-  getKibanaIndex() {
+  @Column({
+    default: false,
+  })
+  is_registered: boolean;
+
+  getKibanaIndex(role: Role) {
     if (this.root_admin) {
       return '*';
     }
-
-    // HACK: We should figure out which role based on which org they're trying to view.
-    const role = this.roles[0];
     return `${role.org.index_prefix}-${role.index_prefix}`;
   }
 
-  getKibanaUserClaim() {
-    const role = this.roles[0];
+  getKibanaUserClaim(role: Role) {
     return `org${role.org.id}-role${role.id}`;
   }
 
-  getKibanaRoles() {
-    // HACK: This should be based on their permissions somehow.
-    return 'kibana_admin';
+  getKibanaRoles(role: Role) {
+    if (role.can_manage_dashboards) {
+      return 'kibana_admin';
+    }
+    return 'kibana_ro_strict';
+
   }
 
 }
--- a/server/app.ts
+++ b/server/app.ts
@@ -10,7 +10,7 @@ import kibanaProxy from './kibana';
 import kibanaDashboard from './kibana/dashboard';
 import database from './sqldb';
 import config from './config';
-import { requireUserAuth } from './auth';
+import { requireOrgAccess, requireUserAuth } from './auth';
 import { errorHandler } from './util/error-handler';
 
 database.then(() => {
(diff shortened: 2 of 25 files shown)


server/app.ts
```diff
@@ -43,8 +43,12 @@ app.get('/heartbeat', (req: Request, res: Response) => {
 });
 
 app.use('/api', apiRoutes);
-app.use('/dashboard', kibanaDashboard);
-app.use(config.kibana.appPath, kibanaProxy);
+app.use('/dashboard',
+  requireOrgAccess,
+  kibanaDashboard);
+app.use(config.kibana.appPath,
```
larskendall, line 49: Same as above

server/auth/index.ts
```diff
@@ -38,31 +41,76 @@ export async function requireUserAuth(req: AuthRequest, res: Response, next: Nex
...
+      } else {
+        throw new NotFoundError('Organization was not found.');
       }
     }
+  }
+  if (req.appOrg) {
+    return next();
+  }
+  throw new ForbiddenError('User does not have sufficient privileges to perform this action.');
+}
```
larskendall, line 107: This function is looking a little clumpy and tough to read to me. Could you break things up with some white space? Maybe after lines 79, 82, 94, 102, and 105.

server/auth/index.ts
```diff
@@ -38,31 +41,76 @@ export async function requireUserAuth(req: AuthRequest, res: Response, next: Nex
...
+export async function requireOrgAccess(req: any, res: Response, next: NextFunction) {
+  let orgId = -1;
+  if (req.params.orgId) {
+    orgId = parseInt(req.params.orgId);
+  } else if (req.query.orgId) {
+    orgId = parseInt(req.query.orgId);
+  } else if (req.cookies.hasOwnProperty('orgId')) {
+    orgId = parseInt(req.cookies.orgId);
+  }
+  if (orgId === -1) {
```
larskendall, line 80: Right now it looks like someone could pass in a negative number or some text for `orgId` and it would make it past the error check...

I think think this might be a little tighter if `orgId` is initially undefined or null and we switch the first check to `if (orgId == null)`, then we add another check `if (isNaN(orgId) || orgId < 0)` and throw a different "Invalid org id" error in that case.

server/config/index.ts
```diff
@@ -37,6 +37,8 @@ if (process.env.NODE_ENV === 'development') {
 type Config = typeof baseConfig
   & Partial<typeof devConfig>
   & Partial<typeof prodConfig>
-  & Partial<typeof testConfig>
+  & Partial<typeof testConfig>;
 
-export default config;
+const configExport = config;
+
+export default configExport;
```
larskendall, line 44: Do these lines do anything different than what was there? It seems like the `configExport` variable is redundant.

src/actions/userActions.ts
```diff
@@ -42,6 +46,13 @@ export namespace User {
       type = Logout.type;
     }
 
+    export class ChangeOrg {
+      static type = 'CHANGE_ORG';
```
larskendall, line 50: It's best to prefix redux actions to avoid naming clashes and just to make debugging a little easier.

```suggestion
      static type = 'USER_CHANGE_ORG';
```

src/actions/userActions.ts
```diff
@@ -60,4 +71,10 @@ export namespace User {
       ...new Actions.Logout(),
     });
   };
+
+  export const changeOrg = (orgId: number) => (dispatch: Dispatch<Actions.ChangeOrg>) => {
+    dispatch({
+      ...new Actions.ChangeOrg({ orgId })
```
larskendall, line 77: Just an eslint thing.

```suggestion
      ...new Actions.ChangeOrg({ orgId }),
```

server/api/user/user.model.ts
```diff
@@ -45,24 +49,28 @@ export class User extends BaseEntity {
...
   }
 
-  getKibanaRoles() {
-    // HACK: This should be based on their permissions somehow.
-    return 'kibana_admin';
+  getKibanaRoles(role: Role) {
+    if (role.can_manage_dashboards) {
+      return 'kibana_admin';
+    }
+    return 'kibana_ro';
```
larskendall, line 72:
```suggestion
    return 'kibana_ro_strict';
```

larskendall (changes requested):
It's all working very nicely. Just a few minor details in the code, and the following UI issues I noticed...

- Changing orgs while in the Roster page doesn't refresh the roster data.
- Org dropdown underline hover color shouldn't be black (I think we should just remove the underline actually)
- We should probably show the org name even if you're only in one org.

This might be a task for later... but we need some way of saving your UI state for when you come back from Kibana or refresh. Right now, depending on which org you have selected, you might come back from Kibana or refresh and be on a different org suddenly. We should probably make this a separate task though, since we'll want to save other UI state stuff as well, like whether you have the sidebar collapsed or not.

PR #2, opened 2020-09-24
Add user registration.

larskendall:


src/actions/userActions.ts
```diff
@@ -77,4 +56,56 @@ export namespace User {
...
+    });
+  };
+}
+
+export interface UserData {
+  edipi: string,
+  first_name: string,
+  last_name: string,
+  phone: string,
+  email: string,
```
jdgarrett, line 82: service?

src/components/pages/UserRegistrationPage/UserRegistrationPage.tsx
```diff
@@ -0,0 +1,153 @@
...
+                id="phone"
+                label="Phone number"
+                required
+                onChange={e => handleInputChange('phone', e)}
+              />
+            </div>
+            <div>
+              <TextField
+                id="organization"
+                label="Organization"
```
jdgarrett, line 119: This should probably say Service to avoid confusion with other app terminology.

src/components/pages/UserRegistrationPage/UserRegistrationPage.tsx
```diff
@@ -0,0 +1,153 @@
...
+                label="EDIPI"
+                value={user.edipi}
+                required
+                disabled
+              />
+              <TextField
+                id="phone"
+                label="Phone number"
+                required
+                onChange={e => handleInputChange('phone', e)}
```
jdgarrett, line 113: Can we add validation on the phone number to make sure we have a consistent format?

src/reducers/userReducer.ts
```diff
@@ -1,100 +1,39 @@
...
-  activeRole: UserRole | undefined
-}
+import { User, UserData } from '../actions/userActions';
 
 export const userInitialState: UserState = {
+  edipi: '',
+  firstName: '',
+  lastName: '',
+  phone: '',
+  email: '',
```
jdgarrett, line 8: Missing service.

src/components/pages/UserRegistrationPage/UserRegistrationPage.tsx
larskendall: Yeah, I didn't see anything built-in, but it looks like this package will work well. I'll see if I can add it easily enough.

https://www.npmjs.com/package/material-ui-phone-number

PR #18, opened 2020-10-07
Add loading spinners

trevermock:
**Here's a high Level of what was added:**

Added page loading spinners to the below pages. The page loading spinner keeps track of state using Redux.
* groups
* roster
* roles
* users

Added a custom button with a loading spinner and then added that button to the below existing buttons. They all keep state locally, except the Create Account button which i used Redux (not sure if that's necessary or not, let me know and I can change it if needed).
* Create Account Button (User Registration page)
* Request Access button (Groups page)
* Finalize Approval button (Users page dialogue)
* Deny Request button (Users page)
* Save Role Button (Role Management page Dialogue)
* Delete Role Button (Role Management page)

### DIFF
--- a/src/actions/user.actions.ts
+++ b/src/actions/user.actions.ts
@@ -34,6 +34,7 @@ export namespace User {
         userData: ApiUser
       }) {}
     }
+
   }
 
   export const login = () => async (dispatch: Dispatch<Actions.Login>) => {
--- a/src/components/pages/groups-page/groups-page.tsx
+++ b/src/components/pages/groups-page/groups-page.tsx
@@ -6,20 +6,23 @@ import { MailOutline, PersonAdd, MoreVert } from '@material-ui/icons';
 import { Alert } from '@material-ui/lab';
 import axios, { AxiosResponse } from 'axios';
 import React, { useEffect, useState } from 'react';
-import { useSelector } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { UserState } from '../../../reducers/user.reducer';
 import { AppState } from '../../../store';
 import { StatusChip } from '../../status-chip/status-chip';
 import useStyles from './groups-page.styles';
 import { ApiAccessRequest, ApiOrg } from '../../../models/api-response';
+import { AppFrame } from '../../../actions/app-frame.actions';
+import { ButtonWithSpinner } from '../../buttons/button-with-spinner';
 
 export const GroupsPage = () => {
   const classes = useStyles();
   const user = useSelector<AppState, UserState>(state => state.user);
-  const [isLoading, setIsLoading] = useState(true);
+  const dispatch = useDispatch();
   const [isAlertVisible, setIsAlertVisible] = useState(!user.activeRole);
   const [isInfoCardVisible, setIsInfoCardVisible] = useState(true);
   const [accessRequests, setAccessRequests] = useState([] as ApiAccessRequest[]);
+  const [accessRequestsLoading, setAccessRequestsLoading] = useState({} as ({[orgId: number]: boolean}));
   const [allOrgs, setAllOrgs] = useState([] as ApiOrg[]);
   const [requestAccessOrgs, setRequestAccessOrgs] = useState([] as ApiOrg[]);
   const [myOrgMenuAnchor, setMyOrgMenuAnchor] = useState(null as HTMLElement | null);
@@ -39,7 +42,15 @@ export const GroupsPage = () => {
     requestAccess(org).then();
   }
 
+  function updateAccessRequestsLoading(org: ApiOrg, isLoading: boolean) {
+    setAccessRequestsLoading({
+      ...accessRequestsLoading,
+      [org.id]: isLoading,
+    });
+  }
+
   async function requestAccess(org: ApiOrg) {
+    updateAccessRequestsLoading(org, true);
     const response = await axios.post(`api/access-request/${org.id}`) as AxiosResponse<ApiAccessRequest>;
     const newRequest = response.data;
 
@@ -53,6 +64,7 @@ export const GroupsPage = () => {
     }
 
     setAccessRequests(accessRequestsUpdated);
+    updateAccessRequestsLoading(org, false);
   }
 
   async function cancelRequest(org: MyOrg) {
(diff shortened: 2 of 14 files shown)


src/actions/user.actions.ts
```diff
@@ -34,6 +34,15 @@ export namespace User {
...
     }
+
+    export class SetRegisterLoading {
+      static type = 'USER_SET_REGISTER_LOADING';
+      type = SetRegisterLoading.type;
+      constructor(public payload: {
+        isRegisterLoading: boolean
+      }) {}
+    }
+
```
trevermock, line 45: Not sure if this info is worth using redux for or not. If not, I can switch it over to not using redux.

src/components/pages/groups-page/groups-page.tsx
```diff
@@ -39,7 +42,14 @@ export const GroupsPage = () => {
     requestAccess(org).then();
   }
 
+  function updateAccessRequestsLoading(org: ApiOrg, isLoading: boolean) {
+    const accessRequestsLoadingClone = { ...accessRequestsLoading };
+    accessRequestsLoadingClone[org.id] = isLoading;
+    setAccessRequestsLoading(accessRequestsLoadingClone);
+  }
```
trevermock, line 49: If I'm understanding React right, I think I should make a clone before I update the state on an object like I did here. With that said, not completely sure if this is necessary or not.

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -99,11 +100,17 @@ export const RosterPage = () => {
...
+
+    dispatch(AppFrame.setPageLoading(true));
+
     fetch(`api/roster/${orgId}/count`).then(async response => {
       const countResponse = (await response.json()) as CountResponse;
       setRosterSize(countResponse.count);
       await handleChangePage(null, 0);
     });
+
+    dispatch(AppFrame.setPageLoading(false));
```
jdgarrett, line 112: This function needs to be updated to be async and instead use `axios.get`.  As is, the page loading will be set to `false` before the asynchronous API request completes.

src/actions/user.actions.ts
jdgarrett: Yes, this seems like it should be a state variable in the registration page.

trevermock: Done. This now uses a state variable.

src/components/pages/roster-page/roster-page.tsx
trevermock: Good catch. Updated and tested this. Should be working correctly now with an axios.get.

larskendall (changes requested):
This is all looking really good, with just a few little things I left comments for.

There was one bug I noticed where the buttons have changed size in the Users page. I'm guessing it's something related to the styling on the wrapper elements in the ButtonWithSpinner component. Here's a screenshot of the bug...

Screen Shot 2020-10-09 at 10 48 53 AM

Aside from that, we'll definitely eventually want to put all of the `setXxxxLoading(false)` calls into `finally {}` blocks, so that the loading spinners hide even if a request fails. But I think it's fine if we do that in a separate PR, since it will involve adding better error handling in general.

src/components/pages/user-registration-page/user-registration-page.tsx
```diff
@@ -31,8 +33,10 @@ export const UserRegistrationPage = () => {
     });
   }
 
-  function handleCreateAccountClick() {
-    dispatch(User.register(inputData));
+  async function handleCreateAccountClick() {
+    setRegisterUserLoading(true);
+    await dispatch(User.register(inputData));
+    setRegisterUserLoading(false);
```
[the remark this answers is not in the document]
trevermock: Yeah, I went back and double checked it to be sure. It seems to work from what I can tell.

src/components/pages/users-page/users-page.tsx
```diff
@@ -16,41 +16,54 @@ import {
...
   const classes = useStyles();
+  const dispatch = useDispatch();
 
   const [formDisabled, setFormDisabled] = useState(false);
   const [activeAccessRequest, setActiveAccessRequest] = useState<AccessRequestRow | undefined>();
   const [selectedRole, setSelectedRole] = useState<number>(0);
   const [availableRoles, setAvailableRoles] = useState<ApiRole[]>([]);
   const [userRows, setUserRows] = useState<ApiUser[]>([]);
   const [accessRequests, setAccessRequests] = useState<AccessRequestRow[]>([]);
+  const [finalizeApprovalLoading, setFinalizeApprovalLoading] = useState<boolean>(false);
```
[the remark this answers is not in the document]
trevermock: Removed the `Boolean` type there

src/components/pages/users-page/users-page.tsx
```diff
@@ -16,41 +16,54 @@ import {
...
     setUserRows(users);
     setAccessRequests(requests);
     setAvailableRoles(roles);
+    dispatch(AppFrame.setPageLoading(false));
   }, [orgId]);
 
+  function updateDenyRequestLoading(rowId: number, isLoading: boolean) {
+    const denyRequestsLoadingClone = { ...denyRequestsLoading };
+    denyRequestsLoadingClone[rowId] = isLoading;
+    setDenyRequestsLoading(denyRequestsLoadingClone);
```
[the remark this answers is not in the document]
trevermock: done

src/components/pages/users-page/users-page.tsx
```diff
@@ -16,41 +16,54 @@ import {
...
+  const dispatch = useDispatch();
 
   const [formDisabled, setFormDisabled] = useState(false);
   const [activeAccessRequest, setActiveAccessRequest] = useState<AccessRequestRow | undefined>();
   const [selectedRole, setSelectedRole] = useState<number>(0);
   const [availableRoles, setAvailableRoles] = useState<ApiRole[]>([]);
   const [userRows, setUserRows] = useState<ApiUser[]>([]);
   const [accessRequests, setAccessRequests] = useState<AccessRequestRow[]>([]);
+  const [finalizeApprovalLoading, setFinalizeApprovalLoading] = useState<boolean>(false);
+  const [denyRequestsLoading, setDenyRequestsLoading] = useState({} as ({[rowId: number]: boolean}));
```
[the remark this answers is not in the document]
trevermock: Typescript threw an error on this this, as we discussed. I left this as-is.

PR #30, opened 2020-10-23
Update workspace templates and fix index patterns

jdgarrett:


larskendall (changes requested):
Looks good to me. Just that one little renaming thing is all I'd suggest.

PR #31, opened 2020-10-23
Sort roster columns

trevermock:
I branched off of the error handling stuff I did, which branched off of the roster enhancement stuff, so we'll want to merge both of those PR's before reviewing this one.

Also, I realize this stuff might be redundant as of the stuff Lars is working on. There were a lot of ways to go about this and I'm not sure if this way fits with the direction you guys are going or not. If you can use it, great! If not though, no worries!

With that said, here's a few notes on what I did:
- Added an optional parameter to the api/roster/:orgId endpoint 'sort_by' which takes the columnName.sortOrder (example: edipi.asc) for how to sort the resulting data.
- On the server end, I just changed the 'orderBy' part of the query to use those optional params. Right now it defaults to edipi.asc if the optional 'sort_by' param isn't passed.
- I'm assuming that the camelCase column name on the front-end will correctly translate to the snake_case column name stored in the database. I'm not sure the correct way to handle that, so that part might have to be given a bit more thought.

I'll squash commits and rebase after the error handling stuff is merged into main, assuming you guys want these changes.


jdgarrett:
This functionality has been implemented on a separate PR.

PR #33, opened 2020-10-26
Table custom columns content.

larskendall:


src/components/pages/roster-page/roster-page.tsx
```diff
@@ -147,9 +72,33 @@ export const RosterPage = () => {
...
+  //
 
-  function handleFileInputChange(e: ChangeEvent<HTMLInputElement>) {
+  const handleChangePage = (event: MouseEvent<HTMLButtonElement> | null, pageNew: number) => {
+    setPage(pageNew);
+  };
+
+  const handleChangeRowsPerPage = (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+    setRowsPerPage(parseInt(event.target.value));
+    setPage(0);
```
jdgarrett, line 98: Can we calculate the page based on which page the first row of the table is going to be on?  For example, if they were on page 6 showing 10 rows per page, they should be on page 2 if they switch to 50 rows per page, or page 3 if they switch to 25 rows per page.

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -169,20 +118,9 @@ export const RosterPage = () => {
           message: `Successfully uploaded ${count} roster entries.`,
           onClose: () => { setAlertDialogProps({ open: false }); },
         });
-        initializeTable();
+        initializeTable().then();
```
jdgarrett, line 121: Can this be awaited?

src/components/tables/table-custom-columns-content.tsx
```diff
@@ -0,0 +1,82 @@
...
+            {columns.map(column => (
+              <TableCell key={`${column.name}-${row.edipi}`}>
+                {row[column.name]}
+              </TableCell>
+            ))}
+
+            {showActions() && (
+              <TableCell className={classes.tableButtons}>
+                <Button
+                  className={classes.editRosterEntryButton}
```
jdgarrett, line 62: Minor nitpick, but I would remove roster from these names.

src/components/tables/table-custom-columns-content.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+          {showActions() && (
+            <TableCell />
+          )}
+        </TableRow>
+      </TableHead>
+
+      <TableBody>
+        {rows.map(row => (
+          <TableRow key={`${row.edipi}`}>
```
jdgarrett, line 52: We can't assume that `edipi` will exist in the row.  In fact, EDIPI will not be present for non-PII views.

src/components/tables/table-custom-columns-content.tsx
```diff
@@ -0,0 +1,82 @@
...
+            {columns.map(column => (
+              <TableCell key={`${column.name}-${row.edipi}`}>
```
jdgarrett, line 54: `edipi` reference.

jdgarrett (changes requested):

src/components/pages/roster-page/roster-page.tsx
larskendall: Yeah I think that would be better UX for sure.

src/components/tables/table-custom-columns-content.tsx
larskendall: Oops

server/api/roster/roster.controller.ts
```diff
@@ -164,13 +164,12 @@ class RosterController {
       })
       .getRawMany<RosterEntryData>();
 
-    res.json(roster);
-  }
-
-  async getRosterCount(req: ApiRequest<OrgParam>, res: Response) {
-    const count = await (await queryAllowedRoster(req.appOrg!, req.appRole!)).getCount();
+    const totalRowsCount = await (await queryAllowedRoster(req.appOrg!, req.appRole!)).getCount();
```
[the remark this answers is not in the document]
larskendall: Hmm this seems like an edge case that might not be worth complicating the code with. That would only be the situation on the very last page of the roster right? I'd rather avoid any early optimizations that aren't going to make an obvious impact in performance.

PR #38, opened 2020-11-09
Add user management

AnthonyWhitaker:
User management per Brian's mockups

### DIFF
--- a/server/api/user/user.controller.ts
+++ b/server/api/user/user.controller.ts
@@ -110,9 +110,10 @@ class UserController {
       user.roles = [];
     }
 
-    const orgRole = user.roles.find(userRole => userRole.org!.id === org.id);
-    if (orgRole) {
-      throw new BadRequestError('The user already has a role in the organization.');
+    const orgRoleIndex = user.roles.findIndex(userRole => userRole.org!.id === org.id);
+
+    if (orgRoleIndex >= 0) {
+      user.roles.splice(orgRoleIndex, 1)
     }
 
     user.roles.push(role);
@@ -187,6 +188,37 @@ class UserController {
     const removedUser = await user.remove();
 
     res.json(removedUser);
+  }
+
+  async removeUser(req: ApiRequest<OrgEdipiParams>, res: Response) {
+    if (!req.appOrg) {
+      throw new NotFoundError('Organization was not found.');
+    }
+
+    const userEDIPI = req.params.edipi;
+
+    const user = await User.findOne({
+      relations: ['roles'],
+      where: {
+        edipi: userEDIPI,
+      },
+      join: {
+        alias: 'user',
+        leftJoinAndSelect: {
+          roles: 'user.roles',
+          org: 'roles.org',
+        },
+      },
+    });
+
+    const roleIndex = (user?.roles ?? []).findIndex((userRole) => userRole.org!.id === req.appOrg!.id);
+
+    if (roleIndex !== -1) {
+      user!.roles!.splice(roleIndex, 1);
+      res.json(await user!.save())
+    } else {
+      res.json({})
+    }
   }
 
   async getAccessRequests(req: ApiRequest, res: Response) {
--- a/src/components/pages/users-page/users-page.tsx
+++ b/src/components/pages/users-page/users-page.tsx
@@ -1,6 +1,8 @@
 import {
   Button,
   Container,
+  Menu,
+  MenuItem,
   Paper,
   Table,
   TableBody,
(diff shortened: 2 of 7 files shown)


server/api/user/user.controller.ts
```diff
@@ -189,6 +190,37 @@ class UserController {
     res.json(removedUser);
   }
 
+  async removeUser(req: ApiRequest<OrgEdipiParams>, res: Response) {
+    if (!req.appOrg) {
```
jdgarrett, line 194: These checks still exist throughout the API, but they are no longer necessary, you can remove them from controller functions whenever you see them.  This check is being handled in the `requireOrgAccess` middleware.

server/api/user/user.controller.ts
```diff
@@ -189,6 +190,37 @@ class UserController {
     res.json(removedUser);
   }
 
+  async removeUser(req: ApiRequest<OrgEdipiParams>, res: Response) {
```
jdgarrett, line 193: Maybe these should be a bit more descriptive, such as `removeUserFromGroup` and `addUserToGroup`.

server/api/user/user.controller.ts
```diff
@@ -189,6 +190,37 @@ class UserController {
...
+      join: {
+        alias: 'user',
+        leftJoinAndSelect: {
+          roles: 'user.roles',
+          org: 'roles.org',
+        },
+      },
+    });
+
+    const roleIndex = (user?.roles ?? []).findIndex((userRole) => userRole.org!.id === req.appOrg!.id);
```
jdgarrett, line 214: I would throw a not found error if the user is `undefined`.  This would allow you to remove the `?` and `!` when accessing it and add some extra safety to the endpoint.

src/components/pages/users-page/users-page.tsx
```diff
@@ -10,129 +12,326 @@ import {
...
+  availableRoles: ApiRole[]
+  confirmButtonText: string
+  loading: boolean
+  onCancel?: () => void
+  onChange: (selectedRole: ApiRole) => void
+  open?: boolean
+  user?: ApiUser
+};
+
+const RoleSelectionDialog: React.FunctionComponent<RoleSelectionDialogProps> = ({
```
jdgarrett, line 65: Can you move this dialog to a separate file?  We have been doing that for all of the other pages and it would be nice to keep things consistent.

src/components/pages/users-page/users-page.tsx
```diff
@@ -10,129 +12,326 @@ import {
...
+  availableRoles,
+  confirmButtonText,
+  loading,
+  onCancel,
+  onChange,
+  open,
+  user,
+}) => {
+  const classes = useStyles();
+  const [selectedRole, setSelectedRole] = useState<number>(0);
```
jdgarrett, line 75: I believe `number` is implicit from the default and does not need to be specified.

src/components/pages/users-page/users-page.tsx
```diff
@@ -10,129 +12,326 @@ import {
...
     await initializeTable();
+    updateDenyRequestLoading(id, false);
   }
 
-  const handleAlertClose = () => {
-    setAlert({ open: false, message: '', title: '' });
-  };
-
-  useEffect(() => { initializeTable().then(); }, [initializeTable]);
+  useEffect(() => { initializeTable(); }, [initializeTable]);
```
jdgarrett, line 334: We have been using `.then()` for these to avoid IDE warnings about unhandled promises.

src/components/pages/users-page/users-page.tsx
```diff
@@ -227,134 +438,71 @@ export const UsersPage = () => {
...
+            </ButtonWithSpinner>
+          </DialogActions>
         </DialogContent>
-        <DialogActions>
-          <Button onClick={handleAlertClose} autoFocus>
-            OK
-          </Button>
-        </DialogActions>
       </Dialog>
+      <RoleSelectionDialog
```
jdgarrett, line 480: Is there a way that these could be re-used?  If all of the logic is moved to the role selection dialog in a separate file, then you could do something similar to the other dialogs where you could indicate whether or not it's an edit or an add as a property, which might simplify it a bit.

jdgarrett (changes requested):
Looks good overall, just a few suggestions for keeping things consistent with the rest of the codebase.

server/api/user/user.controller.ts
AnthonyWhitaker: Awesome. Removing...

server/api/user/user.controller.ts
AnthonyWhitaker: I could see value in doing it that way. The only push back I have is that the REST standard states that DELETE operations should be idempotent. I tend to find it more painful to diverge from the standard as the app grows and becomes more robust. For example, the axios-retry library adheres to the standard and will therefore attempt to retry transient failures for DELETE operations. How set are you on your recommended change? Maybe we should discuss on Slack/Meet?

src/components/pages/users-page/users-page.tsx
AnthonyWhitaker: Yeah... sorry the laziness here. :/

I knew this was going to get another refactor with the "assign role dialog" changes so I wasn't too worried about it being in the same file. It's already pulled out in the "assign role dialog" branch that I'll be pushing in a few.

src/components/pages/users-page/users-page.tsx
AnthonyWhitaker: Strange. I don't think I've seen those warnings. Maybe we should sync on how to get my environment set up the same as yours? I'll add this back in.

src/components/pages/users-page/users-page.tsx
AnthonyWhitaker: Yeah, I wanted to merge these but didn't for two reasons:
1) the "assign role dialog" task was going to touch them again
2) a larger conversation about embracing redux

PR #52, opened 2020-12-01
Error handling

DaveJackson-PE:
Summary of changes
------------------

This PR covers the requirements for the following Asana task:

> [Better input validation and error messages on roster load](https://app.asana.com/0/1188940305683068/1199145504016473)

List of changes
---------------
NOTE: this PR refactors use of ```<AlertDialog>``` and corresponding duplicate code in ```<RosterPage>```

- Added EDIPI validation - 10 digit integer allowing leading zeros
- Added ```value.trim().length``` check in addition to ```value.length``` to ```canSave()```
- Implemented ```<ErrorBoundary>``` component and ```useErrorHandler()``` hook

What code can others reuse?
---------------------------

- ```<ErrorBoundary>``` 
- ```useErrorHandler()```

Should this pull request be tested any more or less than usual?
---------------------------------------------------------------
Due to refactoring ```<AlertDialog>``` use in the ```<RosterPage>```, full regression tests should be performed on error handling.

What are general test steps or cases?
-------------------------------------

1. On Roster page, click 'UPLOAD CSV' button 
2. Select an invalid file
3. Verify the following error dialog:

Roster-Upload-CSV-Error

4. Dismiss the dialog and the list is active again, allowing the user to recover and continue.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -86,46 +86,41 @@ export const App = () => {
               [classes.contentFaded]: appFrame.isPageLoading,
             })}
           >
-            <ErrorBoundary
-              alertDialogProps={alertDialogProps}
-              setAlertDialogProps={setAlertDialogProps}
-            >
-              <Switch>
-                <Route path="/home">
-                  <HomePage />
-                </Route>
-                <Route path="/units">
-                  <UnitsPage />
-                </Route>
-                <Route path="/roster">
-                  <RosterPage />
-                </Route>
-                <Route path="/roster-columns">
-                  <RosterColumnsPage />
-                </Route>
-                <Route path="/workspaces">
-                  <WorkspacesPage />
-                </Route>
-                <Route path="/roles">
-                  <RoleManagementPage />
-                </Route>
-                <Route path="/users">
-                  <UsersPage />
-                </Route>
-                <Route path="/groups">
-                  <GroupsPage />
-                </Route>
-                <Route path="/settings">
-                  <SettingsPage />
-                </Route>
-                <Route path="/muster">
-                  <MusterPage />
-                </Route>
-                <Route path="/*">
-                  <Redirect to="/home" />
-                </Route>
-              </Switch>
-            </ErrorBoundary>
+            <Switch>
+              <Route path="/home">
+                <HomePage />
+              </Route>
+              <Route path="/units">
+                <UnitsPage />
+              </Route>
+              <Route path="/roster">
+                <RosterPage />
+              </Route>
+              <Route path="/roster-columns">
+                <RosterColumnsPage />
+              </Route>
+              <Route path="/workspaces">
+                <WorkspacesPage />
+              </Route>
+              <Route path="/roles">
+                <RoleManagementPage />
+              </Route>
+              <Route path="/users">
+                <UsersPage />
+              </Route>
+              <Route path="/groups">
+                <GroupsPage />
+              </Route>
+              <Route path="/settings">
+                <SettingsPage />
+              </Route>
+              <Route path="/muster">
+                <MusterPage />
+              </Route>
+              <Route path="/*">
+                <Redirect to="/home" />
+              </Route>
+            </Switch>
           </div>
 
           <div
@@ -144,8 +139,11 @@ export const App = () => {
   }
 
   return (
-    <>
+    <ErrorBoundary
+      alertDialogProps={alertDialogProps}
+      setAlertDialogProps={setAlertDialogProps}
+    >
       {routes()}
-    </>
+    </ErrorBoundary>
   );
 };
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import {
   Switch, Route, Redirect,
@@ -23,12 +23,16 @@ import { WorkspacesPage } from './pages/workspaces-page/workspaces-page';
 import { RosterColumnsPage } from './pages/roster-columns-page/roster-columns-page';
 import { SettingsPage } from './pages/settings-page/settings-page';
 import { UnitsPage } from './pages/units-page/units-page';
+import { ErrorBoundary } from './error-boundary/error-boundary';
+import { AlertDialogProps } from './alert-dialog/alert-dialog';
 
 export const App = () => {
   const user = useSelector<AppState, UserState>(state => state.user);
   const appFrame = useSelector<AppState, AppFrameState>(state => state.appFrame);
   const dispatch = useDispatch();
   const classes = useStyles();
+
+  const [alertDialogProps, setAlertDialogProps] = useState<AlertDialogProps>({ open: false });
 
   useEffect(() => {
     dispatch(User.login());
@@ -82,41 +86,46 @@ export const App = () => {
               [classes.contentFaded]: appFrame.isPageLoading,
             })}
           >
-            <Switch>
-              <Route path="/home">
-                <HomePage />
-              </Route>
-              <Route path="/units">
-                <UnitsPage />
-              </Route>
-              <Route path="/roster">
-                <RosterPage />
-              </Route>
-              <Route path="/roster-columns">
-                <RosterColumnsPage />
-              </Route>
-              <Route path="/workspaces">
-                <WorkspacesPage />
-              </Route>
-              <Route path="/roles">
-                <RoleManagementPage />
-              </Route>
-              <Route path="/users">
-                <UsersPage />
-              </Route>
-              <Route path="/groups">
-                <GroupsPage />
-              </Route>
-              <Route path="/settings">
-                <SettingsPage />
-              </Route>
-              <Route path="/muster">
-                <MusterPage />
-              </Route>
-              <Route path="/*">
-                <Redirect to="/home" />
-              </Route>
-            </Switch>
+            <ErrorBoundary
+              alertDialogProps={alertDialogProps}
+              setAlertDialogProps={setAlertDialogProps}
+            >
+              <Switch>
+                <Route path="/home">
+                  <HomePage />
+                </Route>
+                <Route path="/units">
+                  <UnitsPage />
+                </Route>
+                <Route path="/roster">
+                  <RosterPage />
+                </Route>
+                <Route path="/roster-columns">
+                  <RosterColumnsPage />
+                </Route>
+                <Route path="/workspaces">
+                  <WorkspacesPage />
+                </Route>
+                <Route path="/roles">
+                  <RoleManagementPage />
+                </Route>
+                <Route path="/users">
+                  <UsersPage />
+                </Route>
+                <Route path="/groups">
+                  <GroupsPage />
+                </Route>
+                <Route path="/settings">
+                  <SettingsPage />
+                </Route>
+                <Route path="/muster">
+                  <MusterPage />
+                </Route>
+                <Route path="/*">
+                  <Redirect to="/home" />
+                </Route>
+              </Switch>
+            </ErrorBoundary>
           </div>
 
           <div
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -23,8 +23,7 @@ import { WorkspacesPage } from './pages/workspaces-page/workspaces-page';
 import { RosterColumnsPage } from './pages/roster-columns-page/roster-columns-page';
 import { SettingsPage } from './pages/settings-page/settings-page';
 import { UnitsPage } from './pages/units-page/units-page';
-import { ErrorBoundary } from './error-boundary/error-boundary';
-import { AlertDialogProps } from './alert-dialog/alert-dialog';
+import { ErrorBoundary, ErrorDialogProps } from './error-boundary/error-boundary';
 
 export const App = () => {
   const user = useSelector<AppState, UserState>(state => state.user);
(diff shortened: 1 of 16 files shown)


src/components/pages/roster-page/edit-roster-entry-dialog.tsx
```diff
@@ -125,14 +125,25 @@ export const EditRosterEntryDialog = (props: EditRosterEntryDialogProps) => {
...
         switch (column.type) {
           case ApiRosterColumnType.String:
           case ApiRosterColumnType.Date:
-          case ApiRosterColumnType.DateTime:
-            if ((rosterEntry[column.name] as string).length === 0) {
+          case ApiRosterColumnType.DateTime: {
+            if (value.length === 0 || value.trim().length === 0) {
               result = false;
+            } else if (column.name === 'edipi') {
+              const chars = value.split('');
```
jdgarrett, line 137: This can be validated with the regex `/^[0-9]{10}$/g`

DaveJackson-PE: Ha! Yes it can, thanks!

DaveJackson-PE: Done

DaveJackson-PE:
Any other questions, comments or concerns @jdgarrett @larskendall @AnthonyWhitaker @rogelioii ? Just wondering if this looks good to merge yet...

src/components/error-boundary/error-boundary.tsx
```diff
@@ -0,0 +1,88 @@
...
+export interface AlertErrorProps {
+  message?: string,
+  alertTitle?: string,
+  error?: Error,
+  info?: ErrorInfo,
+}
+
+export class AlertError extends Error {
+  readonly alertTitle?: string;
+  readonly error?: Error;
```
larskendall, line 15: This seems a little odd to have an Error type that also has an `error` member. Maybe this should follow the pattern for deriving errors used in `/server/util/error-types.ts`.

src/components/error-boundary/error-boundary.tsx
```diff
@@ -0,0 +1,88 @@
...
+export interface ErrorBoundaryProps {
+  children?: ReactNode,
+  alertTitle?: string,
+  errorMessage?: string,
+  alertDialogProps: AlertDialogProps,
+  setAlertDialogProps: (props: AlertDialogProps) => void,
+}
+
+export interface ErrorBoundaryState {
+  hasError: boolean,
```
larskendall, line 36: Can `hasError` be removed? It looks like it's redundant since `error` is optional and can be checked to determine the same thing.

src/components/error-boundary/error-boundary.tsx
```diff
@@ -0,0 +1,88 @@
...
+    this.setState({
+      hasError: true,
+      error: err,
+    });
+
+    this.props.setAlertDialogProps({
+      open: true,
+      title: this.state.error?.alertTitle ?? 'Error',
+      message,
+      onClose: () => { this.props.setAlertDialogProps({ open: false }); },
```
larskendall, line 73: Should we clear the error here as well when the dialog is closed?

src/components/error-boundary/error-boundary.tsx
```diff
@@ -0,0 +1,88 @@
...
+    if (this.state.hasError) {
+      return (
+        <>
+          {this.props.children}
+          <AlertDialog {...this.props.alertDialogProps} />
+        </>
+      );
+    }
+    return this.props.children;
+  }
```
larskendall, line 87: This block might be a little bit cleaner like this...
```jsx
      return (
        <>
          {this.props.children}
          {this.state.error && (
            <AlertDialog {...this.props.alertDialogProps} />
          )}
        </>
      );
```

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -78,7 +80,10 @@ export const RosterPage = () => {
   const [rowsPerPage, setRowsPerPage] = useState(10);
   const [unitNameMap, setUnitNameMap] = useState<{[key: string]: string}>({});
   const [selectedRosterEntry, setSelectedRosterEntry] = useState<ApiRosterEntry>();
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [alertDialogProps, setAlertDialogProps] = useState<AlertDialogProps>({ open: false });
```
larskendall, line 85: It looks like the success alerts are gone, which is why `alertDialogProps` was unused. I guess they probably got stripped out with the error dialog and you forgot to add them back in.

It should never be necessary to disable that eslint rule on a piece of React state like that, since an unused variable there is a sign that something is broken or unnecessary.

larskendall (changes requested):
The error catching is working nicely! However, it looks like the success alert dialogs in the Roster page are missing now. So those will need to be added back in. Aside from that, just a couple minor things in the code that I left comments for.

There's one possible development workflow issue that I noticed. I'm getting the full page React error from anything that triggers an error dialog, with the error dialog shown behind it. It's not a huge deal, since it only shows in development, but I could see it becoming a nuisance during development always having to close it *and* the error dialog. Plus, that full page error seems like it's designed more to signal unhandled exceptions that need to be addressed in the code. It seems like we need some way to specify whether to treat an error as expected or not, and only show that full page dev error for unexpected errors. Can you think of any elegant ways to do that?

src/components/app.tsx
```diff
@@ -82,41 +86,46 @@ export const App = () => {
...
-                <SettingsPage />
-              </Route>
-              <Route path="/muster">
-                <MusterPage />
-              </Route>
-              <Route path="/*">
-                <Redirect to="/home" />
-              </Route>
-            </Switch>
+            <ErrorBoundary
```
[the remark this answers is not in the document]
DaveJackson-PE: Good point! That sounds like a good plan. I'm on it...

src/components/error-boundary/error-boundary.tsx
DaveJackson-PE: Definitely!

src/components/error-boundary/error-boundary.tsx
DaveJackson-PE: Oh yeah... I think I had that on my TODO list before I went to bed this morning and forgot to get it in before submitting the PR

src/components/error-boundary/error-boundary.tsx
DaveJackson-PE: Good call. I'm on it...

src/components/pages/roster-page/roster-page.tsx
DaveJackson-PE: Yikes! That's not good... I'll make sure those are fixed and take another pass through the changes to see if any similar problems are lurking.

src/components/app.tsx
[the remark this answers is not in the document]
DaveJackson-PE: Committed. Working on the others...

src/components/error-boundary/error-boundary.tsx
DaveJackson-PE: I like it! Before I get too far into this, I'm thinking something along these lines:

```tsx
export class AlertError extends Error {
  constructor(
    message: string,
    readonly type: AlertErrorType,
    readonly title?: string,
    readonly isException: boolean = false,
  ) {
    super(message);
  }
}

export class RosterColumnError extends AlertError {
  constructor() {
    super('Failed to get roster column info', 'RosterColumn', 'Get Roster Column Info');
    Error.captureStackTrace(this, RosterColumnError);
  }
}

export class RosterUploadError extends AlertError {
  constructor(message: string) {
    super(`${message}. Please verify the roster data.`, 'RosterUpload', 'Upload Error');
    Error.captureStackTrace(this, RosterUploadError);
  }
}

export type AlertErrorType = (
  'RosterColumn' |
  'RosterUpload'
);
```

Then, in the roster-page.tsx file, we'd have something like this:

```tsx
  const initializeRosterColumnInfo = useCallback(async () => {
    try {
      const infos = (await axios.get(`api/roster/${orgId}/info`)).data as ApiRosterColumnInfo[];
      const unitColumnWithInfos = [unitColumn, ...infos];
      setRosterColumnInfos(unitColumnWithInfos);
      if (initialLoad.current) {
        setVisibleColumns(unitColumnWithInfos.slice(0, maxNumColumnsToShow));
      }
    } catch (error) {
      handleError(new RosterColumnError());
    }
  }, [orgId, handleError]);
```

Also, note that the `isException` property can be used to distinguish between expected errors and unexpected exceptions.

larskendall: Something just occurred to me... rather than specifying an `isException` bool, can we use `if (error instanceof AlertError) {...}` to determine whether it should show an alert or be treated as an unhandled exception? I was thinking a bool originally as well, but I think it would be even better if the type could be used. That's similar to how the server is working, where regular Error objects are treated as internal server errors.

I'd maybe opt for not getting too specific with the error object types unless the same one is used lots of times. But it could be a nice pattern to establish for when that is the case.

src/components/pages/roster-page/roster-page.tsx
DaveJackson-PE: I resolved this by adding the `<AlertDialog>` back to the page. I created a new `<ErrorDialog>` component for the `<ErrorBoundary>` so success and error dialog components can co-exist.

src/components/error-boundary/error-boundary.tsx
DaveJackson-PE: Definitely a cleaner, simpler approach. Fixed in the latest commit.

src/components/error-boundary/error-boundary.tsx
DaveJackson-PE: Resolved in the latest commit.

src/components/error-boundary/error-boundary.tsx
DaveJackson-PE: Resolved in the latest commit

src/components/error-boundary/error-boundary.tsx
DaveJackson-PE: Resolved in the latest commit

DaveJackson-PE:
Yeah, the React dev error overlay is definitely a pain. Not sure off the top of my head how to solve that elegantly, but I'll dig into it and see what I can find.

larskendall:
@DaveJackson-PE I looked into this issue with the React dev error a bit, and it appears to be essentially impossible to remove it without doing some very sketchy things like cloning and re-adding the `html` element in order to remove the event listeners. The only other way to disable it is to eject create-react-app, which I'd really rather avoid.

I'm thinking that dev error page is going to cause a lot of annoyance in the future, especially if we get errors in something like input, where I think the page would potentially re-display on every keystroke. The more I've read about it, the more I think it actually is meant pretty specifically for unhandled exceptions, since it actually remounts the entire component tree when it gets hit.

So I'm wondering if we should just ditch the error boundary altogether, and instead add a dialog action/reducer to handle this via redux. That would potentially have more utility as well, since we could also streamline basic error dialogs that way. With that approach, I think we would just have a single `AlertDialog` component (or maybe a more specialized component like `DialogUtility` that contains an `AlertDialog`) at the root of the `App` hierarchy, and it would watch some redux state like `state.dialog.alert` to see if it should be displayed.

What are your thoughts?

DaveJackson-PE:
> @DaveJackson-PE I looked into this issue with the React dev error a bit, and it appears to be essentially impossible to remove it without doing some very sketchy things like cloning and re-adding the `html` element in order to remove the event listeners. The only other way to disable it is to eject create-react-app, which I'd really rather avoid.
> 
> I'm thinking that dev error page is going to cause a lot of annoyance in the future, especially if we get some kind of expected error occurring repeatedly. The more I've read about it, the more I think it actually is meant pretty specifically for unhandled exceptions, since it apparently remounts the entire component tree when it gets hit.
> 
> So I'm wondering if we should just ditch the error boundary altogether, and instead add a dialog action/reducer to handle this via redux. That would potentially have more utility as well, since we could also streamline basic alert dialogs that way. With that approach, I think we would just have a single `AlertDialog` component (or maybe a more specialized component like `DialogUtility` that contains an `AlertDialog`) near the root of the app hierarchy, and it would watch some redux state like `state.dialog.alert` to see if it should be displayed.
> 
> What are your thoughts?

Oh crap! That's not good... But, I definitely think we should keep the `<ErrorBoundary>` for exception handling only, since it allows the app to recover gracefully. I believe it's only remounting the entire component tree because of it's position at the root. I haven't tested this yet, but I believe if we have an `<ErrorBoundary>` at the page level, only that page would be remounted. So, it's probably fine for a last-ditch recovery mechanism at the root, but I'm guessing it's most efficient for recovery at lower levels of modularity. Since this Asana issue is `<RosterPage>`-specific, maybe we should add an `<ErrorBoundary>` at the page-level.

I like the idea of a singleton modal `<AlertDialog>` or derivative near the root. And, we might also eventually consider something similar for `Toast` notifications/messages if we're using those. But, I believe these are beyond the scope of this original Asana ticket, so I'd like to just add new tickets to the backlog for these enhancements. Unless, of course, anyone thinks the `<AlertDialog>` piece of the puzzle is within scope. If not, I'm inclined to restore the original `setAlertDialogProps(...)` calls for now. In my opinion, the enhanced validation already committed, as well as the `<ErrorBoundary>` components at the root and roster page level satisfy the requirements for this ticket.

@larskendall @jdgarrett @rogelioii does this sound like a reasonable plan?

larskendall:
Yeah that sounds like a good plan to me. Error boundaries definitely still have utility for unhandled exceptions to prevent the UI from ending up in a corrupted state, so it's good to have them in there as well.

I'd say let's just make that dialog utility stuff a high priority tech debt item, since it will save us a lot of time going forward and keep the code a lot cleaner.

Go ahead and squash/merge this when you're ready and we can tackle that other stuff separately.

DaveJackson-PE:
Sounds good, thanks! To simplify the merge and code review, I created a new branch and opened PR #56 with just the validation and ```<ErrorBoundary>``` changes.

PR #53, opened 2020-12-01
Add API endpoints for finding muster windows

jdgarrett:


DaveJackson-PE (approved):
Looks good to me!

server/api/muster/muster.controller.ts
```diff
@@ -277,6 +281,135 @@ class MusterController {
...
+      let firstWindowStart: number = 0;
+      let lastWindowStart: number = 0;
+      for (let day = DaysOfTheWeek.Sunday; day <= DaysOfTheWeek.Saturday; day = nextDay(day)) {
+        const end = current + durationSeconds;
+        if (dayIsIn(day, muster.days)) {
+          if (firstWindowStart === 0) {
+            firstWindowStart = current;
+          }
+          lastWindowStart = current;
+          const distanceToWindow = timestamp > end ? timestamp - end : (timestamp < current ? timestamp - current : 0);
```
larskendall, line 371: This is pretty tough to read. It also looks very similar to the other `distanceToWindow` bit below. Can we break it into function sort of like this?

```typescript
  function getDistanceToWindow(start: number, end: number, time: number) {
    if (time > end) {
      return time - end;
    }
    if (time < start) {
      return time - start;
    }
    return 0;
  }
```

src/components/pages/units-page/edit-unit-dialog.tsx
```diff
@@ -99,6 +107,9 @@ export const EditUnitDialog = (props: EditUnitDialogProps) => {
       configuration[index].timezone = timezone;
     }
     setMusterConfiguration(configuration);
+    if (errorMessage) {
+      setErrorMessage(null);
+    }
```
larskendall, line 112: I think this could be wrapped in a `resetErrorMessage()` function.

larskendall (commented):
Looks good to me! Just a few suggestions to possibly make things a little cleaner.

PR #54, opened 2020-12-02
Export API

larskendall:
You should have access to the Data Export page if you have a role with a workspace assigned to it. The data that gets exported should reflect the workspace's permissions. By default, with no dates set, it will export everything.

Screen Shot 2020-12-02 at 12 25 06 PM


server/elasticsearch/elasticsearch.ts
```diff
@@ -2,3 +2,45 @@ import { Client } from 'elasticsearch';
 import config from '../config';
 
 export default new Client(config.elasticsearch);
+
+export type EsReportSource = {
```
jdgarrett, line 6: I don't think we can make any assumptions about the structure of the data.  We might have to settle for using `any`.  The reason is the structure will be different based on permissions, custom columns, and other things we might add in the future such as muster information.

server/api/org/index.ts
```diff
@@ -39,4 +42,10 @@ router.post(
   controller.addOrg,
 );
 
+router.get(
```
jdgarrett, line 45: It seems a bit awkward to have this endpoint here, maybe a new `/api/export/:orgId` endpoint?

src/components/pages/data-export-page/data-export-page.tsx
```diff
@@ -0,0 +1,190 @@
...
+      } finally {
+        if (!message) {
+          message = response.statusText;
+        }
+      }
+
+      throw new Error(`Failed to export data: ${message}`);
+    }
+
+    const totalHits = parseInt(response.headers.get('total-hits')!);
```
jdgarrett, line 74: I worry about the case where they have a large date range and the number that comes back here is very large (in the millions).  We could potentially set a threshold to prevent the browser from going out of memory and tell the user to split their export into smaller date ranges.

src/components/pages/data-export-page/data-export-page.tsx
```diff
@@ -0,0 +1,190 @@
...
+      chunk = await reader.read();
+    }
+
+    downloadFile(dataStr, 'export', 'csv');
+  }, [orgId, startDate, endDate]);
+
+  const handleExportButtonClick = () => {
+    setIsExportLoading(true);
+
+    runExport()
```
jdgarrett, line 97: can we change this to use `await` in a `try/catch/finally` block to be more consistent with the rest of the codebase?

jdgarrett (changes requested):

server/api/org/index.ts
larskendall: That's how I originally had it, but it seemed like we might end up with only one route in that endpoint (at least for the foreseeable future). Since it's an export of the org data it felt appropriate that it would go in the org route to me. I'm fine with changing it if you feel strongly about it though.

src/components/pages/data-export-page/data-export-page.tsx
larskendall: Whoops, for some reason I was thinking that function couldn't use async.

PR #56, opened 2020-12-03
Roster page validation and error handling

DaveJackson-PE:
Summary of changes
------------------

This PR covers the requirements for the following Asana task:

> [Better input validation and error messages on roster load](https://app.asana.com/0/1188940305683068/1199145504016473)

List of changes
---------------

- Added EDIPI validation - 10 digit integer allowing leading zeros
- Added `value.trim().length` check in addition to `value.length` to `canSave()`
- Implemented `<ErrorBoundary>` component

What code can others reuse?
---------------------------

- `<ErrorBoundary>` 

Should this pull request be tested any more or less than usual?
---------------------------------------------------------------
Usual regression testing on the Roster page for validation and error handling should be sufficient.

What are general test steps or cases?
-------------------------------------

1. On Roster page, click 'UPLOAD CSV' button 
2. Select an invalid file
3. Verify the following error dialog:

Roster-Upload-CSV-Error

4. Dismiss the dialog and the list is active again, allowing the user to recover and continue.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import {
   Switch, Route, Redirect,
@@ -24,12 +24,15 @@ import { WorkspacesPage } from './pages/workspaces-page/workspaces-page';
 import { RosterColumnsPage } from './pages/roster-columns-page/roster-columns-page';
 import { SettingsPage } from './pages/settings-page/settings-page';
 import { UnitsPage } from './pages/units-page/units-page';
+import { ErrorBoundary, ErrorDialogProps } from './error-boundary/error-boundary';
 
 export const App = () => {
   const user = useSelector<AppState, UserState>(state => state.user);
   const appFrame = useSelector<AppState, AppFrameState>(state => state.appFrame);
   const dispatch = useDispatch();
   const classes = useStyles();
+
+  const [errorDialogProps, setErrorDialogProps] = useState<ErrorDialogProps>({ open: false });
 
   useEffect(() => {
     dispatch(User.login());
@@ -139,8 +142,11 @@ export const App = () => {
   }
 
   return (
-    <>
+    <ErrorBoundary
+      errorDialogProps={errorDialogProps}
+      setErrorDialogProps={setErrorDialogProps}
+    >
       {routes()}
-    </>
+    </ErrorBoundary>
   );
 };
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -1,10 +1,10 @@
 import clsx from 'clsx';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import {
   Switch, Route, Redirect,
 } from 'react-router-dom';
-import { CircularProgress } from '@material-ui/core';
+import { Button, CircularProgress } from '@material-ui/core';
 import { User } from '../actions/user.actions';
 import { AppFrameState } from '../reducers/app-frame.reducer';
 import { UserState } from '../reducers/user.reducer';
@@ -24,15 +24,13 @@ import { WorkspacesPage } from './pages/workspaces-page/workspaces-page';
 import { RosterColumnsPage } from './pages/roster-columns-page/roster-columns-page';
 import { SettingsPage } from './pages/settings-page/settings-page';
 import { UnitsPage } from './pages/units-page/units-page';
-import { ErrorBoundary, ErrorDialogProps } from './error-boundary/error-boundary';
+import { ErrorBoundary } from './error-boundary/error-boundary';
 
 export const App = () => {
   const user = useSelector<AppState, UserState>(state => state.user);
   const appFrame = useSelector<AppState, AppFrameState>(state => state.appFrame);
   const dispatch = useDispatch();
   const classes = useStyles();
-
-  const [errorDialogProps, setErrorDialogProps] = useState<ErrorDialogProps>({ open: false });
 
   useEffect(() => {
     dispatch(User.login());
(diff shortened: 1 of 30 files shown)


src/components/pages/roster-page/roster-page.tsx
```diff
@@ -370,152 +372,157 @@ export const RosterPage = () => {
       <Container maxWidth="md">
         <PageHeader title="Roster" />
```
DaveJackson-PE, line 374: Not sure why the GitHub diff engine gets so confused on these minor changes - the diff is much simpler in VS Code. All I did was insert the ```<ErrorBoundary>``` component into the markup which changed the indentation of the markup it wraps...

src/components/pages/roster-page/edit-roster-entry-dialog.tsx
```diff
@@ -119,22 +119,30 @@ export const EditRosterEntryDialog = (props: EditRosterEntryDialogProps) => {
...
-        if (rosterEntry[column.name] == null) {
-          return false;
+        if ((type === ApiRosterColumnType.String
+          || type === ApiRosterColumnType.Date
+          || type === ApiRosterColumnType.DateTime)
+          && (value.length === 0 || value.trim().length === 0)
+        ) {
+          result = false;
+          break;
         }
```
DaveJackson-PE, line 142: I changed the `switch` statement to `if` so we can `break` out of the loop on an invalid value. Otherwise, the loop continues through the remaining values.

larskendall: I don't think we actually even need that `result` variable. It doesn't look like `result` can ever get set to true again once it's set to false, so we can potentially just do `return false` instead of doing `result = false; break;`, and then put `return true` at the end of the function. Since there's no need to break here we could also get away with a `switch`, but the `if` works fine too.

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -90,6 +91,7 @@ export const RosterPage = () => {
   const [filtersOpen, setFiltersOpen] = useState(false);
   const [visibleColumnsMenuOpen, setVisibleColumnsMenuOpen] = useState(false);
   const [applyingFilters, setApplyingFilters] = useState(false);
+  const [errorDialogProps, setErrorDialogProps] = useState<ErrorDialogProps>({ open: false });
```
larskendall, line 94: I think this should be in ErrorBoundary (related to my comment in the ErrorBoundary component).

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -370,152 +372,157 @@ export const RosterPage = () => {
...
-            accept="text/csv"
-            id="raised-button-file"
-            type="file"
-            style={{ display: 'none' }}
-            ref={fileInputRef}
-            onChange={handleFileInputChange}
-          />
-          <label htmlFor="raised-button-file">
-            <Button
+        <ErrorBoundary
```
larskendall, line 375: I wonder if we should just keep ErrorBoundary near the root of the hierarchy for now (similar to how it was before). Is there any reason you can think of that it needs to be this granular at the moment?

src/components/app.tsx
```diff
@@ -135,8 +138,11 @@ export const App = () => {
   }
 
   return (
-    <>
+    <ErrorBoundary
```
larskendall, line 141: There are a couple blocks above (lines 47 and 58) above that should probably also be wrapped for this to be consistent. I think it would probably make the most sense for now to just put the error boundary around `<App />` in `index.tsx`. That will catch all of these different render paths in one place.

src/components/app.tsx
```diff
@@ -23,13 +23,16 @@ import { WorkspacesPage } from './pages/workspaces-page/workspaces-page';
...
 import { UnitsPage } from './pages/units-page/units-page';
+import { ErrorBoundary, ErrorDialogProps } from './error-boundary/error-boundary';
 
 export const App = () => {
   const user = useSelector<AppState, UserState>(state => state.user);
   const appFrame = useSelector<AppState, AppFrameState>(state => state.appFrame);
   const dispatch = useDispatch();
   const classes = useStyles();
 
+  const [errorDialogProps, setErrorDialogProps] = useState<ErrorDialogProps>({ open: false });
```
larskendall, line 34: I think this should be in ErrorBoundary (related to my comment in the ErrorBoundary component).

src/components/error-boundary/error-boundary.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return AlertDialog(props);
+};
+
+
+export interface ErrorBoundaryProps {
+  children?: ReactNode,
+  errorTitle?: string,
+  errorMessage?: string,
+  errorDialogProps: ErrorDialogProps,
+  setErrorDialogProps: (props: ErrorDialogProps) => void,
```
larskendall, line 20: I'm not sure if passing error data through in the props makes much sense. Right now the Roster and App components are declaring `errorDialogProps` and `setErrorDialogProps`, then passing them into the ErrorBoundary but not actually doing anything with them.

Shouldn't the error boundary itself contain that stuff, with the error that comes through populating them? I don't think the ErrorBoundary component should ever receive any props, since it's potentially higher up in the hierarchy than the component that threw the error.

Putting that stuff in the ErrorBoundary component will look a little different than the other components, since it's a class component and they're functional ones using hooks, but you should be able to do all of the same stuff via its internal state without passing it through in the props.

larskendall (changes requested):
Sorry to keep this PR going! I think there are just a couple anti-patterns that I spotted, which I left comments for.

src/components/pages/roster-page/edit-roster-entry-dialog.tsx
DaveJackson-PE: Removed `result` variable and reverted to `switch` in [latest commit](https://github.com/mystatus/covid19-reports/pull/56/commits/2600dfe7db5443f47c8c15a3a3c69fb8c6611e70).

src/components/pages/roster-page/roster-page.tsx
DaveJackson-PE: Sounds good. Removed from `<RosterPage>`

src/components/pages/roster-page/roster-page.tsx
DaveJackson-PE: Agreed

src/components/app.tsx
DaveJackson-PE: Agreed

PR #57, opened 2020-12-04
Set default elasticsearch index on workspace creation

jdgarrett:
This PR also included the following improvements:

- Update elasticsearch documents when a unit name is changed.
- Fix kibana manifest error spam in logs
- Prevent kibana saved objects from being returned to the client inadvertently

### DIFF
--- a/server/api/workspace/workspace.controller.ts
+++ b/server/api/workspace/workspace.controller.ts
@@ -27,7 +27,7 @@ class WorkspaceController {
       },
     });
 
-    res.json(workspaces);
+    res.json(workspaces.map(sanitizeWorkspace));
   }
 
   async getOrgWorkspaceTemplates(req: ApiRequest, res: Response) {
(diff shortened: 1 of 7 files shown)


server/api/workspace/workspace.controller.ts
```diff
@@ -172,11 +172,19 @@ class WorkspaceController {
...
     const updatedWorkspace = await workspace.save();
 
-    res.json(updatedWorkspace);
+    res.json(sanitizeWorkspace(updatedWorkspace));
   }
 
 }
 
+// Don't return kibana saved objects from workspace endpoints
+function sanitizeWorkspace(workspace: Workspace) {
```
larskendall, line 181: Do you think we should just exclude the kibana saved objects by default in every query using this approach?

https://typeorm.io/#/select-query-builder/hidden-columns

larskendall (commented):
It's all looking good to me. Just those two things to consider.

larskendall (approved):
Looks good!

PR #60, opened 2020-12-07
Multiline string and Enum column support

AnthonyWhitaker:


jdgarrett (changes requested):
Looks good to me!  I made a few recommendations.

PR #62, opened 2020-12-10
Affiliation list updates

JWileczek:
Updates the service affiliation list for new user registration to use these values:
[
  'Air Force',
  'Army',
  'Coast Guard',
  'DOD/OSD',
  'DOD/OSD/DDS',
  'DOD/DHA',
  'Marine Corps',
  'Navy',
  'Space Force',
  'Other',
];
Most are additions, but does remove the 'Contractor' affiliation and replaces with 'Other'.
Contains corresponding DB migration updating any existing 'Contractor' users to the new 'Other' affiliation.

### DIFF
--- a/server/migration/1607616094276-ContractorAffiliationUpdate.ts
+++ b/server/migration/1607616094276-ContractorAffiliationUpdate.ts
@@ -1,0 +1,13 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class ContractorAffiliationUpdate1607616094276 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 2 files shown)


server/migration/1607616094276-ContractorAffiliationUpdate.ts
```diff
@@ -0,0 +1,13 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
```
larskendall, line 1: The migration generator doesn't format according to our eslint rules unfortunately, so these have to be formatted manually. Can you just format this file to match the eslint rules?

larskendall (changes requested):
This is looking good! Just some quick formatting in the migration is needed.

PR #65, opened 2020-12-15
Add redux-persist and save selected group to local storage.

larskendall:
Should resolve this issue https://app.asana.com/0/1188940305683068/1199542270563407

If you refresh the app now it should keep you selected on the same group.

I setup redux-persist to automatically backup data to local storage, using the `whitelist` config option to only backup one special reducer for that purpose so that we can carefully control what goes into local storage and avoid security issues.


larskendall:
I'm not quite sure what happened with the package-lock file and why there were so many deletions. Everything still seems to work fine though, so maybe something weird happened in an earlier commit to give us a ton of redundant data in there.

jdgarrett (approved):

PR #66, opened 2020-12-16
Improve UX for sidenav and roster pages

jdgarrett:


server/api/roster/roster.controller.ts
```diff
@@ -409,37 +311,176 @@ class RosterController {
...
+      const updatedRosterEntry = await manager.save(entry);
+      res.json(updatedRosterEntry);
+    });
+  }
+
+}
+
+async function internalSearchRoster(query: GetRosterQuery, org: Org, role: Role, searchParams?: SearchRosterBody) {
+  const limit = (query.limit != null) ? parseInt(query.limit) : 100;
+  const page = (query.page != null) ? parseInt(query.page) : 0;
```
larskendall, line 374: These can potentially be simplified like this...
```typescript
const limit = parseInt(query.limit ?? '100');
const page = parseInt(query.page ?? '0');
```

server/api/roster/roster.controller.ts
```diff
@@ -409,37 +311,176 @@ class RosterController {
...
+function copyRosterEntry(roster: Roster) {
+  const newEntry = new Roster();
+  newEntry.edipi = roster.edipi;
+  newEntry.unit = roster.unit;
+  newEntry.firstName = roster.firstName;
+  newEntry.lastName = roster.lastName;
+  newEntry.startDate = roster.startDate;
+  newEntry.endDate = roster.endDate;
+  newEntry.lastReported = roster.lastReported;
+  newEntry.customColumns = roster.customColumns;
```
larskendall, line 482: I think `roster.customColumns` needs to be deep cloned. Right now it looks like if the new entry's custom column values get modified it will also affect the original entry.

src/components/app-sidenav/app-sidenav.tsx
```diff
@@ -1,31 +1,87 @@
...
-import ListAltIcon from '@material-ui/icons/ListAlt';
-import PeopleIcon from '@material-ui/icons/People';
+import AssignmentIndIcon from '@material-ui/icons/AssignmentInd';
+import SupervisorAccountIcon from '@material-ui/icons/SupervisorAccount';
 import GroupWorkIcon from '@material-ui/icons/GroupWork';
 import SecurityIcon from '@material-ui/icons/Security';
 import ReplyAllIcon from '@material-ui/icons/ReplyAll';
-import AssessmentOutlinedIcon from '@material-ui/icons/AssessmentOutlined';
+import ViewWeekOutlinedIcon from '@material-ui/icons/ViewWeekOutlined';
+import DashboardoutlinedIcon from '@material-ui/icons/Dashboardoutlined';
```
larskendall, line 13:
```suggestion
import DashboardOutlinedIcon from '@material-ui/icons/DashboardOutlined';
```

src/components/buttons/button-set.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import useStyles from './button-set.styles';
+
+type ButtonSetProps = {
+  children?: React.ReactChild | React.ReactChild[],
```
larskendall, line 5: I think this is the official type for this...
```suggestion
  children?: React.ReactNode,
```

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -104,18 +110,21 @@ export const RosterPage = () => {
 
   const reloadTable = useCallback(async () => {
     try {
-
+      let sortParams = '';
+      if (sortState) {
+        sortParams = `&orderBy=${sortState.column}&sortDirection=${sortState.sortDirection}`;
```
larskendall, line 115: I've started using the `params` option in axios requests, which I think makes for cleaner and less error prone code. There's an example of how to use it in `muster-page.tsx` line 146.

larskendall (changes requested):
I got this build error when I ran the branch initially. It looks like it's just a typo in the import in `app-sidebar.tsx`.
```
Cannot find file: 'Dashboardoutlined.js' does not match the corresponding name on disk: './node_modules/@material-ui/icons/DashboardOutlined.js'.
```

There were a few little UI issues I noticed:
- There's a bug in the New Roster Entry dialog right now where you can't hit Save until you switch the Unit.
- The Roster table sort direction icons should be the other way around.
- Muster page Export button is still in old style.
- We should have tooltips on menu items when the sidebar is collapsed (maybe a separate task)
- Roster "Export to CSV" icon is in reverse direction of "Data Export" menu item. If we want them all the same way as the menu item we should probably just make a custom export icon component that has the icon flipped.
- It looks like there are just a couple lines (`roster-page.tsx` 119,120) that could be removed to fix the issue where the Roster table refreshes differently when filters are applied. After playing with it, I think it's better to have the UI act consistently and we can add a loading spinner as a separate task.

larskendall (commented):
This is all looking great, except I think in the Muster page we should keep the Export Data button in the table since it only exports table data. If it sits outside of the table it sort of implies it's going to export everything on the page.

PR #68, opened 2020-12-17
Adding reingest api endpoint to server.

JWileczek:
Adds simple reingest endpoint to api behind manage group role permissions.
Adds aws-sdk to packages, and config settings for reaching out to dynamoDB table, index, and lambda function.

Do we have any concerns exposing these in the server config?


jdgarrett (changes requested):

server/api/reingest/reingest.controller.ts
```diff
@@ -0,0 +1,131 @@
...
+        lambdaInvocationCount++;
+      }
+
+      res.status(201).json({
+        'reingest-count': recordsIngested,
+        'lambda-invoke-count': lambdaInvocationCount
+      });
+    } catch(err) {
+      console.log('Reingest Failed: ', err.message);
+      res.status(500).send(`Error: ${err.message}`);
```
jdgarrett, line 73: We have been avoiding manually defining error responses.  You can throw a `InternalServerError` with the message and it will be handled appropriately.

server/api/reingest/reingest.controller.ts
```diff
@@ -0,0 +1,137 @@
...
+        'lambda-invoke-count': lambdaInvocationCount
+      });
+    } catch(err) {
+      console.log('Reingest Failed: ', err.message);
+      throw new InternalServerError(`Error during reingest request: ${err.message}`);
+    }
+
+  }
+
+  static convertDateParam(date: string|number|Date): number {
```
larskendall, line 80: I don't think there's any eslint rule for this at the moment unfortunately, but we've generally been formatting this sort of thing like this...

```typescript
static convertDateParam(date: string | number | Date): number {
```

I actually kind of like it crunched together like you had it, but we should just keep it consistent for now.

server/api/reingest/reingest.controller.ts
```diff
@@ -0,0 +1,137 @@
...
+        throw new BadRequestError(`Invalid date string provided for reingest: ${date}`);
+      }
+      return convertedDate.getTime();
+    } else if (date instanceof Date) {
+      return date.getTime();
+    }
+    return date
+  }
+
+  static buildPayloadsFromData(data: AWS.DynamoDB.QueryOutput) {
```
larskendall, line 93: I think these static functions should be moved out of the class, since they appear to just be internal helper functions that don't need to be exported.

server/api/reingest/reingest.controller.ts
```diff
@@ -0,0 +1,137 @@
...
+    return {
+      payloads,
+      recordsIngested
+    };
+  }
+}
+
+export interface TimestampData {
+  startTime?: string|number|Date,
+  endTime?: string|number|Date,
```
larskendall, line 129: Same TS formatting thing as above.

server/api/reingest/reingest.controller.ts
```diff
@@ -0,0 +1,137 @@
...
+      payloads.push(payload);
+    }
+    return {
+      payloads,
+      recordsIngested
+    };
+  }
+}
+
+export interface TimestampData {
```
larskendall, line 127: This probably doesn't need to be exported. If it does, then we should move it to a utils file. Otherwise we run the risk of circular imports between controllers.

server/config/index.ts
```diff
@@ -19,6 +19,12 @@ const baseConfig = {
     httpAuth: (process.env.ELASTICSEARCH_USER && process.env.ELASTICSEARCH_PASSWORD) ? `${process.env.ELASTICSEARCH_USER}:${process.env.ELASTICSEARCH_PASSWORD}` : 'kibana:kibana',
     apiVersion: process.env.ELASTICSEARCH_API_VERSION || '6.4',
   },
+
+  dynamo: {
+    symptomTable: process.env.SYMPTOMTABLE || 'dds-phase1-COVID-19-Symptoms',
+    symptomIndex: process.env.SYMPTOMINDEX || 'EDIPI-Timestamp-index',
+    symptomLambda: process.env.SYMPTOMFUNC || 'symptom_message_handler'
```
larskendall, line 26: Can we make these environment variables a little more explicit? Something like this...
```
DYNAMO_SYMPTOM_TABLE
DYNAMO_SYMPTOM_INDEX
DYNAMO_SYMPTOM_LAMBDA
```

larskendall (changes requested):
It's looking good overall! Just a couple minor comments, and then it needs that eslint pass JD mentioned.

larskendall (approved):
Looks good!

PR #72, opened 2021-01-13
Updated dashboard templates / workspaces for Sprint 11&12

rogelioii:
General Dashboard Updates for Sprint 11 & 12 
Includes Non-Reporting Individuals

### DIFF
--- a/server/kibana/kibana-saved-objects.mock.ts
+++ b/server/kibana/kibana-saved-objects.mock.ts
@@ -3,16 +3,406 @@
 export const kibanaSavedObjectsMock = {
   phi: [
     {
-      "_id": "b1195610-e94e-11ea-b481-91d592e28f0d",
+      "_id": "1fe23f40-2440-11eb-aa43-cf97fefcff0e",
+      "_type": "index-pattern",
+      "_source": {
+        "title": "testgroup1-*-phi-*",
+        "timeFieldName": "Timestamp",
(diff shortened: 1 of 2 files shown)


larskendall (changes requested):
The existing Kibana saved object migrations have followed this same pattern of only having an `up` function, but we should really always have a corresponding `down` to revert if necessary. If other parts of the system are relying on the saved objects being in a certain state then things are potentially going to break if we need to rollback the migration.

The weirdness seems to be that the `down` function depends on a previous version of the SO mock to exist, and those mocks are huge. But I think it would be fine to just inline previous versions into the migration file, with them defined at the bottom of the file beneath the migration logic, so that they're out of the way. Alternatively we could version the SO mock files themselves, but I'd rather not introduce that extra clutter since they would only be there to support the migrations.

I think we should also have a tech debt task to go back and do the same thing for other migrations that are missing a `down` step.

What do you think?

larskendall:
After discussing in the standup, I'd say go ahead and ignore the `down` migration issue for now. I liked Anthony's idea of rolling everything up into a new Initial migration, which we can do as a separate task.

rogelioii:
> The existing Kibana saved object migrations have followed this same pattern of only having an `up` function, but we should really always have a corresponding `down` to revert if necessary. If other parts of the system are relying on the saved objects being in a certain state then things are potentially going to break if we need to rollback the migration.
> 
> The weirdness seems to be that the `down` function depends on a previous version of the SO mock to exist, and those mocks are huge. But I think it would be fine to just inline previous versions into the migration file, with them defined at the bottom of the file beneath the migration logic, so that they're out of the way. Alternatively we could version the SO mock files themselves, but I'd rather not introduce that extra clutter since they would only be there to support the migrations.
> 
> I think we should also have a tech debt task to go back and do the same thing for other migrations that are missing a `down` step.
> 
> What do you think?

Agree. And agree on your what we had talked about in standup today that you also mentioned in your comments.  I will not add the down in this PR, but have created https://app.asana.com/0/1188940305683068/1199545614481853 as a follow-up task

larskendall (approved):

PR #74, opened 2021-01-15
Muster backend updates leveraging new ES data structure.

larskendall:
The Muster page should work just as it did before with no apparent change. However, the backend has been completely reworked to use a much simpler and scalable approach leveraging the `Muster.reported` field that now exists in Elasticsearch.

Here are some other things I added or changed that may be worth noting...

- Added a unique index to the Unit model on the `org` and `name` fields, since an org should never have two different units with the exact same name.
- Modified the `object-curly-newline` eslint rule since it was excessively strict, forcing object destructuring with multiple fields to be broken into multiple lines. Now it will just enforce consistency so that, for any particular destructuring operation, you either put all of the fields on one line, or all on their own line.
- Added a `migration-revert.sh` script to be able to easily run the `down` function in the last applied migration. This is really useful in development.
- Added a `requireQuery` function that can be used to easily validate that query fields are set. If any of the listed fields aren't set, it'll throw an error listing them. I opted not to make it a middleware function, because I think it's better to have that validation right where you're working in the actual endpoint function.
  
  The `requireQuery` function also returns the query object, so you can destructure all of the variables you need at the same time if they don't require any more processing (such as `parseInt`).

  We should add something similar for validating the request body as well, but I didn't really want to add an unused function. We can add it next time we need to validate some body data.

### DIFF
--- a/server/util/util.ts
+++ b/server/util/util.ts
@@ -1,5 +1,6 @@
 import { ValueTransformer } from 'typeorm';
 import moment from 'moment';
+import { ApiRequest } from '../api';
 import { BadRequestError } from './error-types';
 
 
@@ -53,14 +54,13 @@ export function matchWildcardString(str: string, pattern: string) {
   return new RegExp(`^${pattern.split('*').map(escapeRegex).join('.*')}$`).test(str);
 }
 
-export type BaseType = (
-  'string' |
-  'number' |
-  'boolean' |
-  'object' |
-  'function' |
-  'undefined'
-);
+export type BaseType =
+  | 'string'
+  | 'number'
+  | 'boolean'
+  | 'object'
+  | 'function'
+  | 'undefined';
 
 export enum DaysOfTheWeek {
   None = 0,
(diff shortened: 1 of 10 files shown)


server/util/util.ts
```diff
@@ -112,3 +112,65 @@ export const timestampColumnTransformer: ValueTransformer = {
...
+  const missing = getMissingKeys(req.query, keys);
+  if (missing.length) {
+    throw new BadRequestError(`Missing required query params: ${missing.join(', ')}`);
+  }
+
+  return req.query;
+}
+
+function getMissingKeys<T extends object>(obj: T, keys: Array<keyof T>) {
+  const missing = [];
```
AnthonyWhitaker, line 168: Purely preference here but - I find this to be more readable.
`return keys.filter(key => obj[key] === undefined);

server/util/util.ts
```diff
@@ -112,3 +112,65 @@ export const timestampColumnTransformer: ValueTransformer = {
...
+  | 'year';
+
+export function getEsTimeInterval(interval: TimeInterval) {
+  switch (interval) {
+    case 'day': return 'd';
+    case 'week': return 'w';
+    case 'month': return 'M';
+    case 'year': return 'y';
+    default:
+      throw new Error(`Unsupported time interval '${interval}'`);
```
AnthonyWhitaker, line 129: Is this possible since TypeScript should enforce one of the above cases?

server/util/util.ts
```diff
@@ -112,3 +112,65 @@ export const timestampColumnTransformer: ValueTransformer = {
...
+    case 'day':
+    case 'week': return 'yyyy-MM-DD';
+    case 'month': return 'yyyy-MM';
+    case 'year': return 'yyyy';
+    default:
+      throw new Error(`Unsupported interval '${interval}'`);
+  }
+}
+
+export function requireQuery<TQuery extends object>(
```
AnthonyWhitaker, line 155: Super knit-picky but this function name is confusing to me. I think I was thinking of the `require` as `import` - since we're in the JS world. What do you think about ensureRequiredFields or validateQueryKeys or verify/assert 
 ... Something

AnthonyWhitaker (approved):

server/util/util.ts
larskendall: Yep, filter is much better. It totally slipped my mind when writing this.

server/util/util.ts
larskendall: There are two eslint rules preventing the removal of this right now. One is just the `default-case` rule that always requires them - which might be ok to disable - but the other is the `consistent-return` rule, which I would be hesitant to disable or set to ignore undefined. As much as TS does try to enforce the value passed in, it's possible an erroneous value might get forced through, so it seems worth having these error checks in place just in case.

PR #76, opened 2021-01-19
Modal refactor

AnthonyWhitaker:
I backed out some of my changes and fast-pathed the rest of my testing to get this in per Roel's request.


src/utility/errors.ts
```diff
@@ -1,10 +1,13 @@
...
+import { isAxiosError } from './axios';
 
-export function formatMessage(error: Error, message: string, defaultErrorMessage = 'Internal Server Error') {
+export function formatMessage(error: Error, message = '', defaultErrorMessage = 'Internal Server Error') {
   let errorMessage = defaultErrorMessage;
-  const axiosError = error as AxiosError;
-  if (axiosError?.response?.data?.errors && axiosError?.response.data.errors.length > 0) {
-    errorMessage = axiosError?.response.data.errors[0].message;
+  if (isAxiosError(error)) {
+    const axiosError = error as AxiosError;
```
larskendall, line 7: This cast shouldn't be necessary here since the type will already be inferred by the `isAxiosError` type guard.

src/utility/errors.ts
```diff
@@ -1,10 +1,13 @@
...
+    if (axiosError?.response?.data?.errors && axiosError?.response.data.errors.length > 0) {
+      errorMessage = axiosError?.response.data.errors[0].message ?? defaultErrorMessage;
+    }
```
larskendall, line 10: I think we could make this a little easier to read.
```suggestion
    const errors = error?.response?.data?.errors
    if (errors && errors.length > 0) {
      errorMessage = errors[0].message ?? defaultErrorMessage;
    }
```

src/actions/modal.actions.ts
```diff
@@ -0,0 +1,35 @@
...
+        open: boolean,
+        title: string,
+      }) {}
+    }
+
+    export class Close {
+      static type = 'MODAL_CLOSE';
+      type = Close.type;
+      constructor(public payload: {
+        button?: string,
```
larskendall, line 21: It doesn't look like this `button` field is used anywhere. I'm guessing it was added to potentially support confirm dialogs, but probably best to remove it for now to avoid confusion.

src/components/pages/muster-page/muster-page.tsx
```diff
@@ -41,11 +41,12 @@ import {
   ApiRosterColumnType,
```
larskendall, line 41: There are a couple linting errors in this file.

src/components/pages/roster-columns-page/roster-columns-page.tsx
```diff
@@ -19,10 +19,11 @@ import useStyles from './roster-columns-page.styles';
 import { UserState } from '../../../reducers/user.reducer';
```
larskendall, line 19: Linting error on line 3. Can you also reformat that import chunk it's in?

larskendall (changes requested):
This is great. Cleans things up a lot. There are just a few tiny cleanup issues.

I also noticed the `modal.styles.ts` file isn't really doing anything, so we might as well just remove it for now.

src/components/pages/muster-page/muster-page.tsx
AnthonyWhitaker: I don't get an error here. Can you let me know what you're seeing?

AnthonyWhitaker: image

larskendall: It looks like the linting errors were fixed. I'm not seeing any now.

larskendall (approved):
Looks good!

PR #77, opened 2021-01-22
Add support for historical roster

jdgarrett:


server/api/unit/unit.controller.ts
```diff
@@ -145,16 +146,20 @@ class UnitController {
...
       .where('u.org_id = :orgId', { orgId: req.appOrg!.id })
       .andWhere('u.id = :unitId', { unitId: req.params.unitId })
-      .andWhere('(roster.end_date IS NULL OR roster.end_date >= to_timestamp(:timestamp))', { timestamp })
-      .andWhere('(roster.start_date IS NULL OR roster.start_date <= to_timestamp(:timestamp))', { timestamp })
-      .getRawMany() as { edipi: string }[];
+      .andWhere('roster.timestamp <= to_timestamp(:timestamp)', { timestamp })
+      .distinctOn(['roster.edipi'])
+      .orderBy('roster.edipi')
+      .addOrderBy('roster.timestamp', 'DESC')
+      .getRawMany() as { edipi: string, changeType: ChangeType }[];
```
larskendall, line 160: This might be another candidate for possible cleanup using TypeORM functions.

https://typeorm.io/#/find-options/advanced-options

server/migration/1611025383465-RosterHistory.ts
```diff
@@ -0,0 +1,118 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`CREATE TYPE "roster_history_change_type_enum" AS ENUM('added', 'changed', 'deleted')`);
+    await queryRunner.query(`CREATE TABLE "roster_history" ("id" SERIAL NOT NULL, "edipi" character varying(10) NOT NULL, "first_name" character varying(100) NOT NULL, "last_name" character varying(100) NOT NULL, "custom_columns" json NOT NULL DEFAULT '{}', "timestamp" TIMESTAMP NOT NULL, "change_type" "roster_history_change_type_enum" NOT NULL DEFAULT 'changed', "unit_id" character varying NOT NULL, "unit_org" integer NOT NULL, CONSTRAINT "PK_1f540dcb3394b6f2f38f1091cc0" PRIMARY KEY ("id"))`);
+    await queryRunner.query(`ALTER TABLE "roster_history" ADD CONSTRAINT "FK_34e597fefbd4b9d5600513023fd" FOREIGN KEY ("unit_id", "unit_org") REFERENCES "unit"("id","org_id") ON DELETE RESTRICT ON UPDATE NO ACTION`);
+
+    const unitRepo = queryRunner.manager.getRepository(Unit);
+    const rosterHistoryRepo = queryRunner.manager.getRepository(RosterHistory);
+    const allUnits = await unitRepo.find({
+      relations: ['org'],
+    });
```
larskendall, line 19: I wonder if even this sort of thing should be a raw query, since in theory relation names and even table names could have changed on the model. Maybe the relations part will work ok since it's specified by string, but it would be worth double checking to make sure TypeORM doesn't reject that before it sends the query.

larskendall (changes requested):
This is looking great. Just a few little thoughts/details as usual.

larskendall (approved):

PR #78, opened 2021-01-22
Default Muster Config

AnthonyWhitaker:
I'm not overly sure how to fully test this - especially since I don't have Elastic/Kibana running.

image
image
image
image


src/components/pages/units-page/units-page.tsx
```diff
@@ -154,71 +184,106 @@ export const UnitsPage = () => {
...
+                            <div key={JSON.stringify({ muster, index })}>
+                              {musterConfigurationToString(muster)}
+                            </div>
+                          ))}
+                        </TableCell>
+                      )}
+                      {unit.musterConfiguration.length === 0 && (
+                        <TableCell className={classes.noMusterConfiguration}>
+                          {defaultMusterConfiguration?.length ? (
+                            <span>* Using default muster requirements</span>
```
jdgarrett, line 254: There needs to be a bit more control over whether or not a unit uses the default muster configuration.  In the edit unit dialog there should be a checkbox for whether or not the unit uses the default (I believe this is in the mockups).  It should be possible for a unit to override the default to have no muster requirement.  My initial thoughts were if the unit had muster requirements set to `null` it would use the default, but if anything was defined then it would use that, so `[]` would mean to override the default with no muster requirements.

jdgarrett (changes requested):
Overall this looks good, it just looks like a bit more work needs to be done on the unit side to properly support overriding defaults.

src/components/pages/units-page/units-page.tsx
AnthonyWhitaker: MMmmmm... I missed that in the original mockup. Thanks, I'll go with the null vs [] approach.

server/api/unit/unit.controller.ts
```diff
@@ -160,13 +161,14 @@ class UnitController {
 
 
 function setUnitFromBody(unit: Unit, body: UnitData) {
+  console.log('setUnitFromBody', { unit, body });
```
jdgarrett, line 164: Left over console log

server/api/muster/muster.controller.ts
```diff
@@ -86,12 +86,13 @@ class MusterController {
       .leftJoinAndSelect('unit.org', 'org')
       .where('json_array_length(unit.muster_configuration) > 0')
```
jdgarrett, line 87: I'm not really sure what the behavior of `json_array_length` is when the value is null, but it seems like the check should be WHERE `(unit.muster_configuration IS NOT NULL AND json_array_length(unit.muster_configuration) >= 0) OR (unit.muster_configuration IS NULL AND json_array_length(org.default_muster_configuration) > 0)`.  It seems like the current logic would fall back to checking the default muster configuration if the length was 0.

jdgarrett (changes requested):
Seeding the database seems to be broken on this branch, it seems like it's not safe to use the model classes within migrations because it can cause issues.  The changes to `Org` has caused some previous migrations to fail when seeding a fresh database.  We probably need to use a locally defined model for any repositories that are accessed in migrations.

server/api/muster/muster.controller.ts
AnthonyWhitaker: The previous logic didn't return any rows that were 0 in length. The new logic aligns with that behavior but then also returns rows that have NULL if there is a default must config. So are you sure we should be using `>=` in that case? I don't fully understand what's happening here - from a business rules point of view. So let me know.

That said, I believe I can write the code to be more clear. I'll take a shot at it.

PR #80, opened 2021-01-25
Separate Unit from Role

DaveJackson-PE:
WIP PR for [Separate Unit From Role](https://app.asana.com/0/1188940305683068/1199542992771591) Asana task.

More details to follow...

### DIFF
--- a/server/api/access-request/access-request.controller.ts
+++ b/server/api/access-request/access-request.controller.ts
@@ -1,5 +1,5 @@
 import { Response } from 'express';
-import { getConnection } from 'typeorm';
+import { getConnection, getManager } from 'typeorm';
 import { BadRequestError, NotFoundError, UnauthorizedError } from '../../util/error-types';
 import { ApiRequest, OrgParam } from '../index';
 import { Org } from '../org/org.model';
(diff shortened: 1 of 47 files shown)


server/auth/index.ts
```diff
@@ -94,11 +100,11 @@ export async function requireOrgAccess(req: any, res: Response, next: NextFuncti
...
   if (orgId && user) {
-    const orgRole = user.roles!.find(role => role.org!.id === orgId);
-    if (orgRole) {
-      req.appOrg = orgRole.org;
-      req.appRole = orgRole;
-      req.appWorkspace = orgRole.workspace;
+    const orgUserRole = user.userRoles.find(userRole => userRole.role.org!.id === orgId);
+    if (orgUserRole) {
+      req.appOrg = orgUserRole.role.org;
+      req.appRole = orgUserRole.role;
```
larskendall, line 106: Here's the core part of what I think will need to be refactored to switch the entire backend over to using `UserRole` instead of `Role`. This should be changed to `req.appUserRole` with its type changed to `UserRole`.

server/ormconfig.ts
```diff
@@ -35,4 +37,5 @@ export const ormConfig: PostgresConnectionOptions = {
   namingStrategy: new SnakeNamingStrategy(),
   synchronize: (process.env.NODE_ENV === 'development' && process.env.SYNC_DATABASE === 'true'),
   logging: false,
+  // logging: true,
```
larskendall, line 40: This probably just slipped through. We should actually make it something like `logging: process.env.TYPEORM_LOGGING ?? false` so that this doesn't happen in the future. Feel free to add that if you'd like.

src/reducers/user.reducer.ts
```diff
@@ -16,39 +16,48 @@ export const userInitialState: UserState = {
   email: '',
   rootAdmin: false,
   isRegistered: false,
-  roles: [],
+  userRoles: [],
   isLoggedIn: false,
 };
 
 export function userReducer(state = userInitialState, action: any): UserState {
+  let reduced = state;
```
larskendall, line 24: I think this may be a bit of an anti-pattern here. Each case in a reducer is meant to be a totally self-contained thing, with an emphasis on as few mutations as possible (preferably none). So sharing any kind of variable between them seems like it could lead to trouble if someone decided to get "clever" with that `reduced` variable. It's best to keep every case completely isolated so that it's easy to reason about them independently.

src/utility/user-utils.ts
```diff
@@ -3,12 +3,13 @@ import { LocalStorageState } from '../reducers/local-storage.reducer';
...
-  const restoredActiveRole = roles?.find(role => role.org?.id === localStorage.orgId);
+  const userRoles = user.userRoles ?? [];
+  const userRole = userRoles.find(ur => ur.role.org?.id === localStorage.orgId);
+  const restoredActiveRole = userRole?.role;
   return {
     ...user,
-    roles,
-    activeRole: restoredActiveRole ?? roles[0],
+    userRoles,
+    activeRole: restoredActiveRole ?? userRoles[0].role,
```
larskendall, line 12: Do you think we should just do this instead? It seems like it's better to have access to the user role throughout the app, since you can access `role` from it if necessary.
```suggestion
    activeRole: restoredActiveRole ?? userRoles[0],
```

larskendall (changes requested):
It looks like `defaultIndexPrefix` is being used in several places where we should be using the user role `indexPrefix`. As far as I can tell, `defaultIndexPrefix` should pretty much never be used beyond assigning the user role's index prefix initially.

Related to this, I think we need to refactor `appRole` to be `appUserRole` (changing its type to `UserRole`), since the user role is really what we're concerned with in virtually every situation. 

I think we can also take the `getUnitFilter` and `getKibana___` functions out of the Role class, since those functions should probably only ever be called on the user role. The Role model seems to essentially be abstract at this point.

Some other general style things...
- I'd say let's just use the name `manager` for those `transactionEntityManager` variables to avoid the verbosity.
- You can potentially simplify those manager calls from 
  `await manager.findOne<User>('User', {` 
  to 
  `await manager.findOne(User, {`
- It looks like there's a lot of code within transaction blocks that doesn't need to be there (such as find queries before/after the core transactional parts). I think we should try to keep only necessary parts like save operations within transactions, just as best practice.

server/api/access-request/access-request.controller.ts
```diff
@@ -126,63 +126,67 @@ class AccessRequestController {
...
-      throw new BadRequestError('User already has a role in the organization');
-    }
+      const userRole = user.userRoles.find(ur => ur.role.org!.id === orgId);
+      if (userRole?.role) {
+        await accessRequest.remove();
+        throw new BadRequestError('User already has a role in the organization');
+      }
 
-    user.roles!.push(role);
+      user.addRole(transactionalEntityManager, role);
```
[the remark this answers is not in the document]
DaveJackson-PE: Fixed in latest commit

DaveJackson-PE:
Thanks for your time and effort on this review @larskendall I believe I've addressed all of the issues you raised, but I'm still going through the transactions to determine what models can be removed. What I've seen so far is that the models I'm referencing through the EntityManager are being updated within the scope of the transaction. It's been a long week and a half, so I'm probably missing something, but I'll pick this up again in the morning.

DaveJackson-PE:
Still making my way through the EntityManager transactions, but I see what you're saying @larskendall  - after a good night's sleep. This round should be finished shortly...

PR #82, opened 2021-01-27
Separate Unit from Role

DaveJackson-PE:
WIP PR for [Separate Unit From Role](https://app.asana.com/0/1188940305683068/1199542992771591) Asana task.

More details to follow...

### DIFF
--- a/server/api/user/user.model.ts
+++ b/server/api/user/user.model.ts
@@ -49,41 +49,74 @@ export class User extends BaseEntity {
   @OneToMany(() => UserRole, userRole => userRole.user)
   userRoles!: UserRole[];
 
-  async addRole(entityManager: EntityManager, role: Role, indexPrefix?: string): Promise<UserRole> {
-    let userRole = await entityManager.findOne(UserRole, {
-      where: {
-        user: this,
-        role,
-      },
-    });
+  async addRole(manager: EntityManager, role: Role, indexPrefix?: string): Promise<User> {
+    let userRole = await this.findUserRole(role);
     if (!userRole) {
-      userRole = entityManager.create<UserRole>('UserRole', {
+      userRole = manager.create<UserRole>('UserRole', {
         user: this,
         role,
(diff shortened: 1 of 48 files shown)


larskendall (changes requested):
Skimming over the code again, it's looking good to me. I'm sure I've driven you crazy with OCD-level requests by now.

However, when I tested the app I noticed an error...

When I try changing a user's role from "Group User" to "Group Admin" I get an internal server error. Here's the error line from the server. It seems to come from within SQL, so there's no good trace unfortunately.

`QueryFailedError: null value in column "user_edipi" violates not-null constraint`

I also got this same error when trying to approve a new user who requested access.

I don't think I had actually tested those things in the previous review, so I can't say whether they were introduced with your most recent changes or not.

As far as I could tell, that was the only issue. I think once it's fixed we can finally wrap this PR up.

DaveJackson-PE:
Not good! I must have just introduced this in my last round of changes. I'll track it down and push the fix ASAP! Thanks.

server/sqldb/seed-dev.ts
```diff
@@ -45,17 +45,17 @@ export default (async function() {
 }());
```
DaveJackson-PE, line 45: In this file, I just renamed `entityManager` to `manager` for consistency with the rest of the code

server/api/user/user.model.ts
```diff
@@ -49,41 +49,74 @@ export class User extends BaseEntity {
...
 
-  async addRole(entityManager: EntityManager, role: Role, indexPrefix?: string): Promise<UserRole> {
-    let userRole = await entityManager.findOne(UserRole, {
-      where: {
-        user: this,
-        role,
-      },
-    });
+  async addRole(manager: EntityManager, role: Role, indexPrefix?: string): Promise<User> {
+    let userRole = await this.findUserRole(role);
```
DaveJackson-PE, line 53: The primary changes to this file were to fix the uninitialized `userRoles` relation, and to `push` and `splice` array entries appropriately.

PR #88, opened 2021-01-29
Fix exception when getting kibana index in user role.

larskendall:
I've tested this locally to make sure it works. If you don't have Kibana setup to test this easily, you can just take a quick glance at the code to make sure it looks ok.


DaveJackson-PE (approved):
Looks good!

PR #91, opened 2021-02-01
Updated for awscliv2

mpaguilar-pe:
The new Jenkins server has awscli v2. These are the recommended changes per https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html#cliv2-migration-ecr-get-login


on-the-edge-4-me (approved):
Changes look good to me

on-the-edge-4-me:
Hold off on merging it until we have fully cut over to deploying with the new jenkins. If we merge this now, we won't be able to deploy from the old jenkins in development. Otherwise, nice work!

on-the-edge-4-me:
Requested change for this. We no longer use the commercial cloud. So we will need to update the aws account in this to the gov cloud.

PR #95, opened 2021-02-07
Finish separating unit from role by allowing a unit to be assigned to users

jdgarrett:


larskendall (changes requested):
The units dropdown is so much nicer than the old filter input. Much much less cryptic. And everything seems to be working great.

In the Users page, we should probably change the "Change Role" menu item to say something like "Edit" instead, since you're doing more than changing role with that dialog now. And the same for the title of that dialog - right now it says "Role Assignment for {Username}" but it should probably just be "Edit {Username}". Or if "Edit" seems too ambiguous maybe something like "Group Settings".

We should have Brian make another pass on the dialogs at some point, since they're starting to feel pretty awkward. Maybe we just need dividers breaking things up conceptually. I'd also kind of like to do away with the two-column thing. Your eyes have to jump all over the place right now to figure out what you're looking at.

larskendall (approved):

PR #97, opened 2021-02-10
Validation for registration page

AnthonyWhitaker:


src/components/pages/user-registration-page/user-registration-page.tsx
```diff
@@ -109,21 +176,35 @@ export const UserRegistrationPage = () => {
...
                 required
-                onChange={e => handleInputChange('phone', e)}
+                {...getInputValidation('phone')}
               />
+              {/* <TextField
+                id="phone"
+                label="Phone number"
+                required
+                onChange={e => handleInputChange('phone', e)}
+              /> */}
```
larskendall, line 196: This looks safe to remove.

larskendall (changes requested):
This is looking nice. I noticed one bug where the Phone Number label stays red even after I've entered a correct phone number. All of the other fields seem to work perfectly though.

What do you think about using libraries for email and phone validation instead of our own functions? It looks like what you added works, but it would be nice to just use what's already out there and battle tested. I was looking at these...

https://www.npmjs.com/package/validator (for email)
https://www.npmjs.com/package/awesome-phonenumber (this will also convert to standard formats)

AnthonyWhitaker:
The phone number component we're using seems to have all the validation we'd need, it just doesn't format very well and I took the lazy way out with the code I have in place. I like the schema oriented validators like Yep or Joi, but I'm all for whatever we want as a team

PR #98, opened 2021-02-11
Removes UnitID from UI

AnthonyWhitaker:


src/components/pages/units-page/edit-unit-dialog.tsx
```diff
@@ -278,17 +275,17 @@ export const EditUnitDialog = (props: EditUnitDialogProps) => {
     }
     setFormDisabled(true);
     const body = {
-      id,
+      id: unit?.id ?? `${name}_${uuidv4()}`.replace(/[^a-z0-9_]/gi, '').toLowerCase(),
```
larskendall, line 278: I don't think this should happen on the client anymore if we can help it. Since it's not an input in the UI anymore, I think we can just generate the id on the backend instead.

We may even be able to remove `Unit.id` altogether. `Unit.org` already has `primary: true` set, and there's currently a unique index using `Unit.org` and `Unit.name`. So I think we may be able to remove that unique index and just use `org` and `name` as a composite id.

src/components/pages/user-registration-page/user-registration-page.tsx
```diff
@@ -26,6 +26,7 @@ function validatePhone(phone: string) {
 }
 
 type UserRegisterDataWithValidation = {
+  /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
larskendall, line 29: It looks like this isn't necessary. I'm not getting any errors when removing it.

src/components/pages/units-page/edit-unit-dialog.tsx
```diff
@@ -55,6 +56,7 @@ interface MusterWindow {
   end: number,
 }
 
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
larskendall, line 59: It looks like this isn't necessary. I'm not getting any errors when removing it.

larskendall (changes requested):

src/components/pages/user-registration-page/user-registration-page.tsx
AnthonyWhitaker: I agree... but for whatever reason - I was getting an error. So weird.

src/components/pages/units-page/edit-unit-dialog.tsx
JWileczek: I agree with Lars we should move this process to generate the id to the backend, but given your comment how these unit ids work into the ES index prefix we should take some more time to think about that change before making any model changes surrounding the Id.
Using UUID would make the chance of id collision minimal, but would impact us down the line with the effort of supporting multiple units. We are going to have to feed the accessible ES indices through a request header to kibana, so the longer the unit id value is the less units at a single time we are going to be able to support.
I think for now we should move this id generation to the backend, and create a ticket to move toward the primary auto-generated ids. Since that will take some more effort with role migrations + reindexing of ES data to match new id format.

JWileczek (changes requested):

jdgarrett:
This is addressed in #107

PR #100, opened 2021-02-12
Roster search fixes for case sensitivity and dates

AnthonyWhitaker:


larskendall (commented):
Looks like it's working well!

I think it would be worth refactoring this `internalSearchRoster` function a bit to try and make it more readable. It's super difficult to follow right now, partly just because it's a huge function, and also because there are now two nested function definitions within it (which is part of what makes it huge). I think just moving those nested function definitions out would make it much easier to follow.

The reducer is really hurting the readability as well. You have to scroll _way_ down to see its starting value, then scroll _way_ back up to follow the logic. I'd just do away with it and try to make it flow top to bottom.

This function should probably also have some comments since it's pretty complicated.

PR #101, opened 2021-02-13
Backend testing framework and initial tests.

larskendall:
You should now be able to run tests with `npm test`. Some of the tests may be a bit rough, since I was moving as fast as I could. And there aren't any error responses that are being tested for yet. But we should have pretty decent coverage to make sure the following sections of the API are at least functional:

`user`
`org`
`access-request`
`role`
`roster` (most endpoints... ran out of time!)

There are also tests for every model I could find that has a relation where some onDelete cascade/restrict/set-null should be occurring.

Mocha is being used for running tests, and chai expect is the assertion handler, with a couple additional chai utilities.

There's one funky change, where I wrapped `app.ts` in a promise in order to await on some things to make sure everything is synced up and ready before tests get kicked off. It shouldn't affect anything else, but let me know if you think there's a better way to solve that.

I added a custom logger, so that we have better control over what shows up in the console when we run tests. We should start using it instead of `console` wherever possible. Sorry for sort of shoehorning that into this PR, but the test output ends up looking like a mess if you don't clean up the logs somehow. It's pretty much the same logger that I added to StatEngine, so it's at least been vetted there a bit.

There were a few bugs that revealed themselves while writing tests, so I went ahead and fixed those so that we have those spots covered and don't have broken tests right out of the gate. They were fairly minor changes though - mostly having to do with missing/incorrect cascades on the models.

I also added a variety of utilities and mocks to aid in writing tests efficiently. Mostly they're for generating data quickly with unique values and seeding the database. There's also a `TestRequest` utility class which is a fairly thin wrapper around axios to be able to easily set things up for tests, such as the base path and the user sending the request. There's also a built-in axios mock adapter for future mocking of external requests.

In order to prepare for any future mocks that might be necessary, I went ahead and added sinon, even though it's not being used by anything yet. I figured we might as well have it ready to go since we'll almost definitely need it at some point.

The Testing section of the readme has some extra details as well if you're interested.


larskendall:
I just tried looping over all of the entities to clear them - instead of deleting each type manually - but it turns out to be _much_ slower (at least 3x) since the entities can be deleted in any particular order that way and have to rely on cascades. So I think I'll just leave it for now until we can find a better solution.

Unfortunately there doesn't seem to be any neat way to simply clear all of the tables out, and I've never even seen a clean way to do that in psql that doesn't involve just dropping and recreating the database.

jkugler:
@larskendall What about using https://www.postgresql.org/docs/9.1/sql-truncate.html

jkugler:
One general comment, but I have not testing experience in TypeScript, so this may be moot. When I've done testing in Ruby and Python, the tests usually live in their own tree off the top level. So, you'd have your src/ or module/ directory, and then a tests/ directory. Is it TS convention to put the spec files right along side the source files?

larskendall:
> @larskendall What about using https://www.postgresql.org/docs/9.1/sql-truncate.html

That's actually what I tried where the tests ended up running much more slowly. I was just looping over all of the tables and truncating them.

larskendall:
> One general comment, but I have no testing experience in TypeScript, so this may be moot. When I've done testing in Ruby and Python, the tests usually live in their own tree off the top level. So, you'd have your src/ or module/ directory, and then a tests/ directory. Is it TS convention to put the spec files right along side the source files?

@jkugler I don't know if there's any particular convention... I've seen it done both ways, so I think it's mostly just personal preference.

I can see how it would be nice to avoid clutter by separating the tests, but to me there's a very important benefit of keeping the tests right next to the files they're testing, which is that it's easier to remember the tests are there and that they need to be kept in sync with the codebase. If they're tucked away in their own folder hierarchy they're much easier to forget about. Not to mention, it's more to think about to remember to keep those hierarchies synced up with each other.

But do you think there's any other major benefit to having them all separated?

One other way I've seen it done is just dropping all of the test files into a test folder adjacent to the files they're testing. That still gets them out of the way for the most part, but keeps them in clear sight as a reminder.

jkugler:
> But do you think there's any other major benefit to having them all separated?

I don't know if I see a benefit other than decluttering as you mention. I think the reason it's done that way in Python is because you import the files which contain the tests, and then find the tests in them. But...even then, you could just import the files with the tests and not the code files...so I don't know.  I suppose whatever the team agrees on.  Either way. :)

server/api/org/org.model.ts
```diff
@@ -44,4 +50,26 @@ export class Org extends BaseEntity {
     default: '[]',
   })
   defaultMusterConfiguration: MusterConfiguration[] = [];
+
+  async getUsers() {
+    const orgUsers: User[] = [];
+
+    const roles = await Role.find({
```
jdgarrett, line 57: This logic is actually flawed as it exposes user roles for other orgs that the caller may not have access to.  Take a look at #104 for how this is fixed.  It would actually be a good test to check that each user in the response has exactly one role for the current org.

server/api/roster/index.ts
```diff
@@ -2,17 +2,18 @@ import express from 'express';
...
 import bodyParser from 'body-parser';
+import { Log } from '../../util/log';
 import controller from './roster.controller';
 import { requireInternalUser, requireOrgAccess, requireRolePermission } from '../../auth';
 
 const rosterUpload = multer({
   storage: multer.diskStorage({
     destination: (req: any, file: any, cb: any) => {
-      console.log('destination');
+      Log.info('destination');
```
jdgarrett, line 12: This seems like a pointless log statement?

server/api/roster/index.ts
```diff
@@ -2,17 +2,18 @@ import express from 'express';
...
       cb(null, path.join(__dirname, 'uploads'));
     },
     filename: (req: any, file: any, cb: any) => {
-      console.log('filename');
+      Log.info('filename');
```
jdgarrett, line 16: Same as above.

jdgarrett (changes requested):
This looks awesome! There were a couple of things that I noticed, but nothing major.

server/run-tests.sh
```diff
@@ -0,0 +1,35 @@
...
+  shift
+done
+
+export NODE_ENV="test"
+export SYNC_DATABASE="false"
+export SQL_DATABASE="dds_test"
+export LOG_LEVEL="$LOG_LEVEL"
+
+# Create test database if it doesn't exist.
+if [ "$(psql -tAc "SELECT 1 FROM pg_database WHERE datname='$SQL_DATABASE'" -h "$SQL_HOST" -p "$SQL_PORT" -U "$SQL_USER")" != '1' ]; then
```
JWileczek, line 21: When attempting to run this PR I ran into some strange output from this command. It created the db successfully the first time, but it doesn't look like this check is entirely correct.
`psql: error: FATAL:  database "jwileczek" does not exist
Creating 'dds_test' database... createdb: error: database creation failed: ERROR:  database "dds_test" already exists
failed to create 'dds_test' database. Aborting...`
It looks like the initial check for the dds_test database isn't always correct. I changed it to use this method of checking if that db exists that works for me:
`if [ "$(psql -lqt | cut -d \| -f 1 | grep -qw $SQL_DATABASE)"];`

JWileczek, line 21: Ran into an interesting problem here. This command failed for me initially because installing postgres from homebrew didn't create a default database under my username. So when running initially I hit the error
`psql: error: FATAL:  database "jwileczek" does not exist`
I ultimately just needed to run `createdb` from the terminal to create a default database for my postgres, but I'm wondering if we should run a drop on the test database beforehand so we create it every time. It wouldn't matter that way if the user has the default user database setup before running tests. It would make us have to run migrations every time, but that may not be bad thing either?
`dropdb $SQL_DATABASE --if-exists`

server/run-tests.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo -n "Creating '$SQL_DATABASE' database... "
+  createdb -h "$SQL_HOST" -p "$SQL_PORT" -U "$SQL_USER" "$SQL_DATABASE" || {
+    echo "failed to create '$SQL_DATABASE' database. Aborting..."
+    exit
+  }
+  echo "success!"
+fi
+
+../migration-run.sh
+npx --silent ts-mocha \
```
JWileczek, line 31: The place of `--silent` applied to this command caused an error when trying to run the first time. Placing it before the command caused this error to throw:
`npx --silent ts-mocha   --project "tsconfig.json"   --config "./.mocharc.yml"   --exit   "./**/*.spec.ts"
 ERROR: You must supply a command.
`
Placing after the `ts-mocha` command seems to fix this issues

JWileczek (commented):
This is looking really great, and I think after addressing a few small issues mentioned in the comments it's in a good spot to bring in. As I was reviewing I was noticing a pattern for the api tests where we create these mock functions to seed data into the database. I'm wondering if we can create some generic interface that models can implement to enforce the ability to mock a backend entity for testing. I notice the format of all the `.mock` files basically falls into:
`seed(requiredFK: Model, customData? Partial<Model>)`
I'm wondering if we can create some `Mockable<T>` interface for each model that defines a seed/mock function. That way we can avoid creating another file for mocking, and wrap it in the model logic. @larskendall @jdgarrett @jkugler Thoughts?

jkugler:
Instead of Mocking, does TypeScript have any support for database fixtures? Importing a file full of SQL would be faster than running a bunch of functions to create records. Or, alternately, have a bunch of data sitting in a data structure that would then be iterated and inserted.

server/api/org/org.model.ts
larskendall: Cool. I rebased and used the latest code for this function.

larskendall:
> Instead of Mocking, does TypeScript have any support for database fixtures? Importing a file full of SQL would be faster than running a bunch of functions to create records. Or, alternately, have a bunch of data sitting in a data structure that would then be iterated and inserted.

@jkugler The upside of using functions is that we can stay within the TypeORM ecosystem, which will keep everything in sync for us and also just simplifies a lot of SQL queries. There doesn't seem to be too much of a performance hit from using the functions, so it should be fine for now. But this is something we can consider as a potential optimization if we start seeing major performance issues with tests down the line. In my experience, though, this approach will scale well to hundreds of tests at least.

larskendall:
> This is looking really great, and I think after addressing a few small issues mentioned in the comments it's in a good spot to bring in. As I was reviewing I was noticing a pattern for the api tests where we create these mock functions to seed data into the database. I'm wondering if we can create some generic interface that models can implement to enforce the ability to mock a backend entity for testing. I notice the format of all the `.mock` files basically falls into:
> `seed(requiredFK: Model, customData? Partial<Model>)`
> I'm wondering if we can create some `Mockable<T>` interface for each model that defines a seed/mock function, or create some class alongside the Model for mocking i.e. `UserMock`. That way we can avoid creating another file for mocking, and wrap it in/with the model logic. @larskendall @jdgarrett @jkugler Thoughts?

Just for the others... after discussing this with John a bit over Slack, we decided that it's best to keep test/mock code and app code separate, and that we'll be on the lookout for ways to enforce mock/seed structure going forward, but that the current setup will work well enough for now.

PR #102, opened 2021-02-16
Add link to onboarding documentation to registration and homepage

JWileczek:


src/components/pages/user-registration-page/user-registration-page.styles.ts
```diff
@@ -59,6 +59,11 @@ export default makeStyles((theme: Theme) => createStyles({
 
     '& *:nth-child(1)': theme.typography.h1,
     '& *:nth-child(2)': theme.typography.subtitle1,
+    '& a:nth-child(1)': {
+      ...theme.typography.subtitle1,
+      fontWeight: theme.typography.fontWeightBold,
+
+    },
```
larskendall, line 66: It looks like the existing styles here weren't quite specific enough, which is why that link was requiring a special override to control its size. Doing this instead should generalize a bit better...
```suggestion
    '& > *:nth-child(1)': theme.typography.h1,
    '& > *:nth-child(2)': theme.typography.subtitle1,
    '& a': {
      fontWeight: theme.typography.fontWeightBold,
    },
```

larskendall (changes requested):
Is there any reason we need the onboarding link to be an environment variable? If it's going to be the same across all environments, and we just want to have it defined in one place, it would probably be better to make something like `urls.ts` under the `utils` folder and drop it in there instead. That might be a better place for it regardless, so that we're not referencing environment variables in lots of places, and so we could default it if we wanted to.

larskendall (approved):

PR #107, opened 2021-02-19
Add support for assigning multiple units to users

jdgarrett:


src/components/pages/roster-page/edit-roster-entry-dialog.tsx
```diff
@@ -82,7 +82,7 @@ export const EditRosterEntryDialog = (props: EditRosterEntryDialogProps) => {
     updateRosterEntryProperty('unit', event.target.value);
   };
 
-  const getUnitName = (unitId: string) => {
+  const getUnitName = (unitId: number) => {
```
JWileczek, line 85: This function always seems to return 'Unknown' when testing locally. Change seems to occur correctly on backend (actually changing unit assigned in roster), but dialog always pops up 'Are you sure you want to move individual to Unknown?' Not sure if it that is a factor of our seeding in local build or not.

JWileczek (changes requested):
Looking mostly correct! Just have a single comment that needs to be addressed/clarified. Also was wondering if/how this will impact our Roster export to CSV functionality, and if we have any tickets made for changes around that if necessary. Looking at the CSV export it looks like it exports unitIds which are just numbers now, and hidden from the frontend, so there is no real way for a user to reassign a unit via CSV without some mental gymnastics to figure out which units align to which IDs. Not real sure if that needs to be addressed in this PR, but we should probably bring it up/make a ticket for it.

src/components/pages/roster-page/edit-roster-entry-dialog.tsx
jdgarrett: This was caused because the incoming unit ID was actually a string due to an incorrect handling of the unit changed event.

jdgarrett:
> Looking mostly correct! Just have a single comment that needs to be addressed/clarified. Also was wondering if/how this will impact our Roster export to CSV functionality, and if we have any tickets made for changes around that if necessary. Looking at the CSV export it looks like it exports unitIds which are just numbers now, and hidden from the frontend, so there is no real way for a user to reassign a unit via CSV without some mental gymnastics to figure out which units align to which IDs. Not real sure if that needs to be addressed in this PR, but we should probably bring it up/make a ticket for it.

I updated the roster CSV export to return unit names instead of IDs.  Import now only works on unit names.

PR #109, opened 2021-02-22
Orphaned record backend

AnthonyWhitaker:


server/api/orphaned-record/orphaned-record.controller.ts
```diff
@@ -0,0 +1,242 @@
...
+      },
+    });
+    if (existingAction) {
+      await existingAction.remove();
+    }
+
+    const orphanedRecords = await OrphanedRecord
+      .createQueryBuilder('orphan')
+      .where('orphan.joinKey=:joinKey', { joinKey: req.params.id })
+      .getMany();
```
larskendall, line 190: Can this use `OrphanedRecord.find()`?

server/api/orphaned-record/orphaned-record.controller.ts
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    // 1231231231;17035813720;unit;1
+
+    const now = Date.now();
+    const orphanedRecordAction = new OrphanedRecordAction();
+    orphanedRecordAction.id = req.params.id;
+    orphanedRecordAction.type = req.body.action;
+    orphanedRecordAction.user = req.appUser;
+    orphanedRecordAction.createdOn = new Date(now);
```
larskendall, line 203: I don't think you need to set `createdOn` since it's already a generated column.

server/api/orphaned-record/orphaned-record-action.model.ts
```diff
@@ -0,0 +1,53 @@
...
+  @CreateDateColumn({
+    type: 'timestamp',
+    transformer: timestampColumnTransformer,
+  })
+  createdOn!: Date;
+
+  @Column({
+    type: 'timestamp',
+    transformer: timestampColumnTransformer,
+    default: () => `now() + interval '1 day'`,
```
larskendall, line 43: It looks like `expiresOn` is being set manually when this model is created. Could we get away with removing the default? I'm just worried we might forget to set it and end up with an unexpected expiration time.

server/api/roster/roster.controller.ts
```diff
@@ -31,6 +31,44 @@ import { CustomColumnValue, RosterColumnInfo, RosterColumnType } from './roster.
 import { UserRole } from '../user/user-role.model';
 import { ChangeType, RosterHistory } from './roster-history.model';
 import { baseRosterColumns } from './roster-entity';
+import { Role } from '../role/role.model';
+
+export async function addRosterEntry(org: Org, role: Role, rosterEntryData: RosterEntryData) {
```
larskendall, line 36: Can you move this to `/server/util/roster-utils.ts`? It also needs to take a `manager` as mentioned above so that it can be used inside transactions.

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -81,6 +82,7 @@ export const RosterPage = () => {
 
   const maxNumColumnsToShow = 5;
 
+  const [orphanedRecords, setOrphanedRecords] = useState<ApiOrphanedRecord[]>([]);
```
larskendall, line 85:
```suggestion
  const [, setOrphanedRecords] = useState<ApiOrphanedRecord[]>([]);
```

larskendall (changes requested):
Sorry for all the comments! There are a couple things that looked like possible bugs (missing `awaits` and some transaction stuff), but most of the comments are just organizational stuff, and trying to stay within the TypeORM ecosystem as much as possible.

On a side note, we really need some tests for this stuff, since it's not clear how to manually test any of this right now. It doesn't need to be part of this PR, but we should consider it a high priority task since this stuff could very easily break without any of us realizing it. I'll make a note to bring it up at the next standup.

In general, we should just get in the habit of adding/updating tests any time backend features are added/updated, since we're pretty much never going to get allocated time to specifically write tests.

PR #117, opened 2021-03-09
Orphaned records

AnthonyWhitaker:
### Overview
image

### Menu
image

### After Claiming a Record
image

### Menu after Claiming (no claim, no ignore)
image

### Ignore Confirmation
image

### Claim Confirmation
image

### Perma Delete Confirmation
image


src/reducers/modal.reducer.ts
```diff
@@ -1,17 +1,53 @@
...
+      const payload = (action as Modal.Actions.Close).payload;
+      if (typeof modalResolverStore[payload.title] === 'function') {
+        modalResolverStore[payload.title](payload.response);
+        delete modalResolverStore[payload.title];
+      }
+    }
+    return next(action);
+  };
+}
+
```
larskendall, line 50: Is this a workaround to avoid saving functions in the Redux store? It looks like this is essentially doing that anyways, just in a hidden store.

What do you think about going ahead and putting the resolve functions in the main store? If we're already essentially doing the same thing I'd rather just keep it simple if we can. I know it's not recommended in Redux, but it's not strictly against the rules. Modals are such a temporary part of the UI that I don't think it would actually hurt anything.

If we do need to do it this way, I'd drop some comments in there at least.

larskendall (changes requested):
- I think the "Add to a Roster" menu option should be "Add to Roster..."

- Maybe the "Assign to Self..." button should just be "Claim...". The dialog that pops up says "Claim Orphaned Record" at the top, so it seems to be the terminology that we're already using.

- Text is a little messed up in Claim dialog
Screen Shot 2021-03-10 at 12 00 11 PM

- Do we need an "Unclaim" option, for when you've claimed something but then changed your mind? It might be bad if you accidentally claim something for a week and then can't undo. It seems like it would be simple enough to drop in and would prevent some awkward situations.

- The buttons on the bottom of Claim and Delete dialogs aren't in the same style as other dialogs. There's normally a different color background behind the buttons, and the buttons aren't quite so close to the bottom.

- The First/Latest Report columns seem like they should have time on the right of the date. It looks backwards to me right now.

- Can we nudge the red orphaned record badge in the sidenav left a couple pixels? It's a little uncomfortably close to the edge.

src/reducers/modal.reducer.ts
AnthonyWhitaker: Putting them in the main store would break serialization. In theory. This pattern is from: https://stackoverflow.com/questions/35325195/should-i-store-function-references-in-redux-store#:~:text=No%2C%20you%20should%20not%20store,of%20hotkeys%20to%20their%20actionCreatorFuncNames.

larskendall (approved):

larskendall:
This looks great. I'll go ahead and merge.

PR #119, opened 2021-03-12
One-Time Muster Windows

AnthonyWhitaker:
# Edit Unit Dialog Changes
image

# First Row Disabled since it's inherited
image

# Now with One-Time super duper reporting
image

# Still can't have overlapped repeating windows
image

# Default Muster Requirements Dialog updated to match
image

# Units page updates to show the unit config plus the simultaneous indication of 'Using default master requirements'
image


server/api/muster/index.ts
```diff
@@ -17,6 +17,13 @@ router.get(
   controller.getNearestMusterWindow,
 );
 
+router.get(
+  '/:orgId/:unitId/open',
```
jdgarrett, line 21: What is this endpoint used for?

jdgarrett (changes requested):
Overall this looks great! I did notice a few logic issues in the nearest/open muster windows functions.  I think tests for these methods would go a long way.  It would be good to set up different muster configurations for different units and test nearest and open window logic for before/during/after.

Also, I wasn't able to see what the `/:orgId/:unitId/open` endpoint was used for.

larskendall (approved):
Looks good! I'm going to forego the usual code deep dive since we're limited on hours right now.

PR #121, opened 2021-03-17
Simplified Oprhaned Records UX

AnthonyWhitaker:
# Removed `Claimed Until` Column
image

# Remove `Claim...` and `Unclaim` menu options
image

# Removed non-Permanent Ignore and updated text
image


larskendall (approved):
Looks good!

PR #125, opened 2021-03-25
Broadcast to symptom tables

AnthonyWhitaker:
Completely untested. Roel made me do it.

### DIFF
--- a/server/util/reingest-utils.ts
+++ b/server/util/reingest-utils.ts
@@ -1,4 +1,4 @@
-import AWS, { AWSError } from 'aws-sdk';
+import AWS from 'aws-sdk';
 import config from '../config';
 import { dateFromString } from './util';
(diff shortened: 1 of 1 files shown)


server/util/reingest-utils.ts
```diff
@@ -143,10 +143,34 @@ export async function reingest(queryInput: AWS.DynamoDB.QueryInput) {
...
+      ...queryInput,
+      TableName: table,
+    });
+    lambdaInvocationCount += result.lambdaInvocationCount;
+    recordsIngested += result.recordsIngested;
+
+    // Avoid additional querying since we've already hit found/processed the record.
+    if (lambdaInvocationCount || recordsIngested) {
+      continue;
+    }
```
larskendall, line 162: This chunk doesn't seem to be doing anything. Is that `continue` mean to be a `break`?

PR #126, opened 2021-03-30
Update new user flow.

larskendall:
Users should now be required to enter information about what they do and who's sponsoring their request to a group.

The approve user flow has also been updated to display all of their answers to the admin.

## Groups Page
Screen Shot 2021-03-30 at 10 53 52 AM

## Users Page
Screen Shot 2021-03-30 at 10 54 58 AM

Screen Shot 2021-03-30 at 10 55 13 AM


server/migration/1617052908638-NewAccessRequestFlow.ts
```diff
@@ -0,0 +1,29 @@
...
+  name = 'NewAccessRequestFlow1617052908638';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "access_request" ADD "what_you_do" character varying array NOT NULL`);
+    await queryRunner.query(`ALTER TABLE "access_request" ADD "sponsor_name" character varying NOT NULL`);
+    await queryRunner.query(`ALTER TABLE "access_request" ADD "sponsor_email" character varying NOT NULL`);
+    await queryRunner.query(`ALTER TABLE "access_request" ADD "sponsor_phone" character varying NOT NULL`);
+    await queryRunner.query(`ALTER TABLE "access_request" ADD "justification" character varying NOT NULL`);
+    await queryRunner.query(`COMMENT ON COLUMN "user_notification_setting"."last_notified_date" IS NULL`);
+    await queryRunner.query(`ALTER TABLE "user_notification_setting" ALTER COLUMN "last_notified_date" SET DEFAULT null`);
```
jdgarrett, line 16: What are these `user_notification_setting` migrations meant to do?  Does it fix the issue with it always wanting to migrate that column?

jdgarrett (changes requested):
Looks good to me!  I do wonder if it should be possible to sign up without a sponsor, but I suppose that's a question for the future.  The only issue I noticed was the units dropdown is not populated correctly when editing a user's role/units.

PR #129, opened 2021-04-01
Allow displayName and name for column heading in csv import

AnthonyWhitaker:
## Resulting UI after importing the two CSVs below
image

### Import with `column.name` headings
image

### Import with `column.displayName` headings
image

### DIFF
--- a/server/api/roster/roster.controller.ts
+++ b/server/api/roster/roster.controller.ts
@@ -195,14 +195,13 @@ class RosterController {
     };
 
     roster.forEach((row, index) => {
-      const units = orgUnits.filter(u => row.unit === u.name);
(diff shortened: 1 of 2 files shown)


AnthonyWhitaker:
## Now with __DODID__
image

## And with __dodid__
image

Tested with edipi, EDIPI, dodid, and DODID
image

image

image

image

jdgarrett (changes requested):
This looks good to me, but as part of this, we should make sure that none of the values of `edipiKey` are valid names for custom columns to make sure there are no conflicts.

AnthonyWhitaker:
Handled by combined PR #131

PR #146, opened 2021-05-13
Sticky Filters and bug fixes

AnthonyWhitaker:


src/hooks/use-deep-equality.ts
```diff
@@ -1,18 +1,38 @@
...
+
+//   const setIt = useCallback((v: SetStateAction<T | undefined>) => {
+//     if (!deepEquals(v, deep)) {
+//       setDeepValue(v);
+//     }
+//   }, [deep, setDeepValue]);
+
+//   return [deep, setIt];
+// }
+
```
larskendall, line 38:
```suggestion
```

src/hooks/use-deep-equality.ts
```diff
@@ -1,18 +1,38 @@
 import {
-  useState, useEffect, SetStateAction, Dispatch,
+  Dispatch, SetStateAction, useCallback, useState,
 } from 'react';
 import deepEquals from 'fast-deep-equal';
 
-export default function useDeepEquality<T>(value?: T | undefined): [T | undefined, Dispatch<SetStateAction<T | undefined>>, T | undefined] {
-  const [deep, setDeepValue] = useState(value);
-  const [candidate, setCandidateValue] = useState(value);
+export default function useDeepEquality<T>(initialValue?: T | undefined): [T | undefined, Dispatch<SetStateAction<T | undefined>>] {
```
larskendall, line 6: I hadn't noticed this hook before. I feel like it might be a little bit of an anti-pattern, since any time you see a `setValue()` sort of function you tend to assume it works a certain way and this breaks that assumption. It would be clearer in the code I think to just see...
```typescript
if (!deepEqual(newValue, value)) { 
  setValue(newValue);
}
```

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -277,7 +291,7 @@ export const RosterPage = () => {
     }
   }, [fetchOrphanedRecords]);
 
-  // Roster reload.
+  // // Roster reload.
```
larskendall, line 294:
```suggestion
  // Roster reload.
```

src/components/query-builder/query-builder.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
```
larskendall, line 1:
```suggestion
import React, { useEffect, useState } from 'react';
```

src/components/query-builder/query-builder.tsx
```diff
@@ -433,14 +435,12 @@ export const QueryBuilder = (props: QueryBuilderProps) => {
...
-        return accum;
-      }, {} as QueryFilterState,
-    );
-
-    setFilterState(query || undefined);
+    const query: QueryFilterState = {};
+    for (const { field, op, value } of queryableRows) {
+      query[field!.name] = { op, value };
+    }
+    console.log('qe', queryableRows, rows, query, filterState);
```
larskendall, line 442:
```suggestion
```

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -79,6 +79,7 @@ import { Modal } from '../../../actions/modal.actions';
 import { UserSelector } from '../../../selectors/user.selector';
```
larskendall, line 79: There are a couple `useEffect()` calls in this file with warnings on the dependencies. Can you just double check those are as they should be and disable the warning with a comment if they look ok?

src/hooks/use-debounced.ts
```diff
@@ -1,11 +1,20 @@
...
-    return () => clearTimeout(timeout);
+    if (isInitialMount.current) {
+      isInitialMount.current = false;
+    } else {
+      const timeout = setTimeout(() => {
+        setDebounced(value);
+      }, delay);
+
+      return () => clearTimeout(timeout);
+    }
```
larskendall, line 17: I think this will alleviate the need for the eslint disable comment...

```suggestion
    }
    
    return () => {};
```

src/reducers/local-storage.reducer.ts
```diff
@@ -51,6 +52,15 @@ export function localStorageReducer(state = localStorageInitialState, action: an
         },
       };
     }
+    case Persist.Actions.Set.type: {
+      const { persistKey, value } = (action as Persist.Actions.Set).payload;
+      console.log('persistAction', persistKey, value);
```
larskendall, line 57:
```suggestion
```

src/reducers/local-storage.reducer.ts
```diff
@@ -51,6 +52,15 @@ export function localStorageReducer(state = localStorageInitialState, action: an
...
+      return {
+        ...state,
+        [persistKey]: action.payload.value,
+      };
+    }
+
```
larskendall, line 63:
```suggestion
```

src/reducers/local-storage.reducer.ts
```diff
@@ -51,6 +52,15 @@ export function localStorageReducer(state = localStorageInitialState, action: an
         },
       };
     }
+    case Persist.Actions.Set.type: {
+      const { persistKey, value } = (action as Persist.Actions.Set).payload;
+      console.log('persistAction', persistKey, value);
+      return {
+        ...state,
+        [persistKey]: action.payload.value,
```
larskendall, line 60:
```suggestion
        [persistKey]: value,
```

src/store.ts
```diff
@@ -29,6 +29,10 @@ import {
   orphanedRecordInitialState,
   orphanedRecordReducer,
 } from './reducers/orphaned-record.reducer';
+// import {
+//   persistInitialState,
+//   persistReducer,
+// } from './reducers/persist.reducer';
```
larskendall, line 35:
```suggestion
```

src/store.ts
```diff
@@ -71,6 +75,7 @@ export const initialState = {
   user: userInitialState,
   workspace: workspaceInitialState,
   localStorage: localStorageInitialState,
+  // persist: persistInitialState,
```
larskendall, line 78:
```suggestion
```

src/store.ts
```diff
@@ -118,7 +123,7 @@ function createPlainObjectMiddleware(): Middleware<AnyAction, AppState> {
 const persistConfig = {
   key: 'root',
   storage,
-  whitelist: ['localStorage'],
+  whitelist: ['localStorage'/* , 'persist' */],
```
larskendall, line 126:
```suggestion
  whitelist: ['localStorage'],
```

src/store.ts
```diff
@@ -137,6 +142,7 @@ export const configureStore = () => {
         user: userReducer,
         workspace: workspaceReducer,
         localStorage: localStorageReducer,
+        // persist: persistReducer,
```
larskendall, line 145:
```suggestion
```

larskendall (changes requested):
It's working well! Just a few ideas about the code and some cleanup notes.

src/hooks/use-persist.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  Dispatch, SetStateAction, useEffect, useRef,
+} from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { Persist } from '../actions/persist.actions';
+import { PersistSelector } from '../selectors/persist.selector';
+
+export default function usePersist<T>(persistKey: string | undefined, [value, setValue]: [T, Dispatch<SetStateAction<T>>]): [T, Dispatch<SetStateAction<T>>] {
```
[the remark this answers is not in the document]
AnthonyWhitaker: That's how I had things at first, but my goal was to make it composable since I needed the useDeepEquality() mechanism. That said, I ended up working around that issue so I could go with the simpler API.

server/util/orphaned-records-utils.ts
```diff
@@ -71,9 +74,8 @@ export function buildVisibleOrphanedRecordResultsQuery(userEdipi: string, orgId:
     .addSelect('orphan.composite_id', 'id')
     .addSelect('roster.unit_id', 'unitId')
     .addSelect('roster.id', 'rosterHistoryId')
-    .orderBy('orphan.count', 'DESC')
-    .addOrderBy('orphan.unit', 'ASC')
-    .addOrderBy('orphan.edipi', 'ASC')
+    .orderBy('orphan.unit', 'ASC')
+    .addOrderBy('MAX(orphan.timestamp)', 'DESC')
```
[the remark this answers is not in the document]
AnthonyWhitaker: This groups by unit first since their workflow is reportedly by unit. And the `MAX()` is there to get the latest timestamp for the current grouped aggregate (the value that shows up as the "Last Reported On" column in the orphaned record UI.).  This really isn't part of the current PR but it was done today with Austin and Bingles based on their feedback from the Lt.

src/components/pages/roster-page/roster-page.tsx
```diff
@@ -139,8 +144,9 @@ export const RosterPage = () => {
   //
   // Effects
   //
-
   const fetchRoster = useCallback(async () => {
+    const fetchRosterId = Date.now() + Math.random();
+    fetchRosterLatestId.current = fetchRosterId;
```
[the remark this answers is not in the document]
AnthonyWhitaker: I'm out for the next 10 days hence the rushing. The calls to the `fetchRoster` function need to work in a `take the last requested result` sort of way. This basically keeps track of the last call to `fetchRoster` and then only updates the state values for the last call... kind of a debounce type of mechanism. It definitely could use a comment or two.

src/hooks/use-deep-equality.ts
AnthonyWhitaker: I'm not sure I agree with the anti-pattern aspect - there are useThrottle, useDebounce, etc... types of hooks, all of which are effectively alterations to the default assumptions provided by useState. 
Personally, I like to encapsulate this type of logic where possible. That said, I do agree that we could just get rid of this hook and use deepEqual since ideally we only have the deep equal need in this particular one-off case. useDeepEqual is testable though... hmm, I don't know. Either way, you've given me something to ponder - I'll think on it some more.

src/hooks/use-debounced.ts
AnthonyWhitaker: Interesting. I had that but I thought "Lars is going to yell at me for not having a comment on this awkward noop line."

src/hooks/use-deep-equality.ts
AnthonyWhitaker: It also protects you from forgetting to wrap your setValue() with the deepEquals check - in those cases where you call setValue from multiple places or pass it to a component as a prop. The solution, in that case, would probably be to use a useCallback (which at that point you're back to the same result as having the useDeepEquality hook...

```javascript
const setValueIfNotDeeplyEqual = useCallback((...) => { 
  if (!deepEqual(newValue, value)) { 
    setValue(newValue);
  }
}, [...]);
```

larskendall: Cool, we can leave it for now. It might be nice at some point to make a quick pass through any of those hooks that are wrappers for `useState` and make them `useState____` just for clarity.

src/hooks/use-debounced.ts
larskendall: Haha yeah it's an odd one. Always nice to avoid disabling eslint when possible though.

src/hooks/use-debounced.ts
```diff
@@ -1,11 +1,22 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 
 export default function useDebounced<T>(value: T, delay: number) {
+  const isInitialMount = useRef(true);
   const [debounced, setDebounced] = useState(value);
 
+  // eslint-disable-next-line consistent-return
```
larskendall, line 7:
```suggestion
```

larskendall (approved):

PR #149, opened 2021-05-21
Roster CSV download/upload null value and date fixes.

larskendall:
https://app.asana.com/0/1188940305683068/1200341565834207

When downloading a roster CSV, you should no longer see "null" as the value for null columns. They should be empty columns instead. And when uploading a roster CSV, empty column values will be treated as null when saved to the database.

I also fixed an issue where the roster CSV template would show different formats for example dates and datetimes. They should now be in a consistent iso format, with example dates stripping off the time part of the iso string.

### DIFF
--- a/server/api/roster/roster-entity.ts
+++ b/server/api/roster/roster-entity.ts
@@ -135,6 +135,9 @@ export abstract class RosterEntity extends BaseEntity {
       value = getOptionalValue(column.displayName, row, 'string');
     }
 
+    // Convert empty strings to null.
+    value = value || null;
+
     this.setColumnValue(column, value);
   }
 
--- a/server/util/roster-utils.ts
+++ b/server/util/roster-utils.ts
@@ -3,7 +3,10 @@ import { Org } from '../api/org/org.model';
 import { Role } from '../api/role/role.model';
 import { RosterHistory } from '../api/roster/roster-history.model';
 import { Roster } from '../api/roster/roster.model';
-import { RosterEntryData } from '../api/roster/roster.types';
+import {
+  RosterColumnInfo,
+  RosterEntryData,
+} from '../api/roster/roster.types';
 import { Unit } from '../api/unit/unit.model';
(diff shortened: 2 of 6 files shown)


server/api/roster/roster-entity.ts
```diff
@@ -135,6 +135,9 @@ export abstract class RosterEntity extends BaseEntity {
       value = getOptionalValue(column.displayName, row, 'string');
     }
 
+    // Convert empty strings to null.
+    value = value || null;
```
AnthonyWhitaker, line 139: Is there a built-in way for the CSV parser to do this for us? Most of them I've used have this as an option

server/util/roster-utils.ts
```diff
@@ -102,3 +105,11 @@ export async function getRosterHistoryForIndividual(edipi: string, unitId: numbe
     .addOrderBy('rh.change_type', 'DESC')
     .getMany();
 }
+
+export function getCsvHeaderForRosterColumn(column: RosterColumnInfo) {
+  return column.displayName;
```
AnthonyWhitaker, line 110: Do we need the mapping for the 'DoD ID' <-> 'edipi' translations and such?

AnthonyWhitaker (approved):
Good stuff. Definitely like the CSV refactorings and the fix for the localeDateString() - I saw that a while ago and it made me uncomfortable. Glad it's addressed.

server/api/roster/roster-entity.ts
larskendall: Looks like there is a way to do that via the csv parser. It might still be good to keep this conversion in place though in case we import from other file types without that parser option.

server/util/roster-utils.ts
larskendall: The `edipi` column's display name is just "DoD ID" now, so this is effectively doing that. From what I could tell, we were mapping to DoD ID everywhere, so we don't seem to need any additional logic for it.

PR #152, opened 2021-06-02
Add unit filter to orphaned records table.

larskendall:
https://app.asana.com/0/1188940305683068/1200394516745667

This is a little hacky at the moment, but Roel was hoping to get it in quickly to demo to the person who requested it.

I tried to make it persistent, but in order to make that work right the units would have to be requested before the orphaned records, and there's also the potential issue of stale persisted data that no longer matches the unit names in the database.

There's a bug right now when you resolve the last orphaned record for a unit, where the Unit dropdown will blank out. It doesn't seem to really break anything though. It's just an empty filter that you can switch off of at that point.

Those aren't super tough issues to fix, but to fix them properly will take a little time. I figure I can make a polish pass tomorrow since I'm out of time for today and Roel wants this by the morning.

Screen Shot 2021-06-02 at 2 44 59 PM

### DIFF
--- a/server/api/orphaned-record/orphaned-record.controller.ts
+++ b/server/api/orphaned-record/orphaned-record.controller.ts
@@ -37,13 +37,14 @@ import { OrphanedRecord } from './orphaned-record.model';
 
 class OrphanedRecordController {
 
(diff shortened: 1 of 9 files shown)


AnthonyWhitaker (approved):
Looks good to me!

PR #155, opened 2021-06-11
Muster page improvements.

larskendall:
https://app.asana.com/0/1188940305683068/1200394516745665
https://app.asana.com/0/1188940305683068/1200416265664999
https://app.asana.com/0/1188940305683068/1200416265665005
https://app.asana.com/0/1188940305683068/1200435628945373

- Flip "muster non-compliance" and "non-muster rate" to "muster compliance" and "muster rate" (except for unit non-compliance trend graphs).

- Show all users in Muster page, even if they're fully compliant. This was a user request, since they were confused about why some users weren't showing up in that page.

- Persist all inputs in Muster page.

- Normalize phone numbers to `(xxx) xxx-xxxx` format across the database, and as they're pulled from Elasticsearch

- Some renaming of functions and variables to make things more consistent/explicit. Most of them involve renaming things from `individuals` to `roster`.

- Update dev seed data to be a little cleaner.

## Screenshots
Screen Shot 2021-06-11 at 10 49 08 AM

Screen Shot 2021-06-11 at 10 35 53 AM

### DIFF
--- a/server/api/muster/index.ts
+++ b/server/api/muster/index.ts
@@ -18,17 +18,17 @@ router.get(
 );
 
 router.get(
-  '/:orgId/individuals',
+  '/:orgId/roster',
   requireOrgAccess,
   requireRolePermission(role => role.canViewMuster),
-  controller.getIndividuals,
+  controller.getMusterRoster,
 );
 
 router.get(
-  '/:orgId/trends',
+  '/:orgId/unit-trends',
   requireOrgAccess,
   requireRolePermission(role => role.canViewMuster),
-  controller.getTrends,
+  controller.getMusterUnitTrends,
 );
 
 export default router;
--- a/server/migration/1623276130329-FormatPhoneNumbers.ts
+++ b/server/migration/1623276130329-FormatPhoneNumbers.ts
@@ -22,7 +22,9 @@ export class FormatPhoneNumbers1623276130329 implements MigrationInterface {
 }
 
 async function updatePhoneNumbers(queryRunner: QueryRunner, formatter: (phone: string) => string) {
+  //
   // Orphaned records.
+  //
   const orphanedRecords = await queryRunner.query(`SELECT "document_id", "phone" FROM "orphaned_record"`) as Array<{
     document_id: string
     phone: string
@@ -33,7 +35,9 @@ async function updatePhoneNumbers(queryRunner: QueryRunner, formatter: (phone: s
     await queryRunner.query(`UPDATE "orphaned_record" SET "phone" = $1 WHERE "document_id" = $2`, [row.phone, row.document_id]);
   }
 
+  //
   // Access requests.
+  //
   const accessRequests = await queryRunner.query(`SELECT "id", "sponsor_phone" FROM "access_request"`) as Array<{
     id: number
     sponsor_phone: string
@@ -44,7 +48,9 @@ async function updatePhoneNumbers(queryRunner: QueryRunner, formatter: (phone: s
     await queryRunner.query(`UPDATE "access_request" SET "sponsor_phone" = $1 WHERE "id" = $2`, [row.sponsor_phone, row.id]);
   }
 
+  //
   // Users.
+  //
   const users = await queryRunner.query(`SELECT "edipi", "phone" FROM "user"`) as Array<{
     edipi: string
     phone: string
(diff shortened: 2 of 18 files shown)


server/api/muster/index.ts
```diff
@@ -18,17 +18,17 @@ router.get(
 );
 
 router.get(
-  '/:orgId/individuals',
+  '/:orgId/roster',
```
AnthonyWhitaker, line 21: I've struggled with this since the beginning. `Roster` refers to both a single `RosterEntry` as well as a collection of them. I was never necessarily a fan of `individual` but it would be nice for us to get consistency around this... good first step here. You rock.

server/migration/1623276130329-FormatPhoneNumbers.ts
```diff
@@ -0,0 +1,57 @@
...
+      const phoneDigits = phone.replace(/\D/g, '');
+      return (phoneDigits.length === 10)
+        ? `${phoneDigits.slice(0, 3)}-${phoneDigits.slice(3, 6)}-${phoneDigits.slice(6)}`
+        : phone;
+    });
+  }
+
+}
+
+async function updatePhoneNumbers(queryRunner: QueryRunner, formatter: (phone: string) => string) {
```
AnthonyWhitaker, line 24: Should this also do Roster? RosterHistory?

server/util/muster-utils.ts
```diff
@@ -92,67 +92,74 @@ export async function getRosterMusterStats(args: {
...
+      rosterStats[edipi].mustersReported = bucket.doc_count;
     }
   }
 
-  // Calculate non-muster percents.
-  for (const edipi of Object.keys(individualStats)) {
-    const data = individualStats[edipi];
-    individualStats[edipi].nonMusterPercent = calcNonMusterPercent(data.mustersReported, data.mustersNotReported);
+  // Calculate muster percents.
+  for (const edipi of Object.keys(rosterStats)) {
```
AnthonyWhitaker, line 146: Might be cleaner as `for (const [edipi, data] = Object.entries(rosterStats)) {`

server/util/muster-utils.ts
```diff
@@ -92,67 +92,74 @@ export async function getRosterMusterStats(args: {
...
     .sort((edipiA, edipiB) => {
       const entryA = rosterEntriesByEdipi[edipiA];
       const entryB = rosterEntriesByEdipi[edipiB];
-      const individualA = individualStats[edipiA];
-      const individualB = individualStats[edipiB];
+      const individualA = rosterStats[edipiA];
+      const individualB = rosterStats[edipiB];
 
-      let diff = individualB.nonMusterPercent - individualA.nonMusterPercent;
+      let diff = individualA.musterPercent - individualB.musterPercent;
```
AnthonyWhitaker, line 162: `const diff`?

server/util/muster-utils.ts
```diff
@@ -92,67 +92,74 @@ export async function getRosterMusterStats(args: {
...
       if (diff === 0) {
```
AnthonyWhitaker, line 163: Should we maybe do an epsilon style comparison here. These are both floating point values and could potentially equal down to 0.000001%, for example. I guess, we could also round them first and compare the rounded values? Either way, I could see a case in the UI where you'd expect them to be alphabetically ordered (because the numerical values both show the same percentage, but they may not be ... Maybe something like 
`if (Math.abs(diff) < 0.001) {` or something?

server/util/string-utils.ts
```diff
@@ -1,9 +1,5 @@
...
-  const phoneDigits = phone.replace(/\D/g, '');
-
-  if (phoneDigits.length === 10) {
-    return `${phoneDigits.slice(0, 3)}-${phoneDigits.slice(3, 6)}-${phoneDigits.slice(6)}`;
-  }
+import { formatNumber } from 'libphonenumber-js';
 
-  return phone;
+export function formatPhoneNumber(phone: string) {
+  return formatNumber(phone, 'US', 'NATIONAL');
```
AnthonyWhitaker, line 4: Ah-mazing

AnthonyWhitaker (approved):

server/api/muster/index.ts
larskendall: That Roster naming has always driven me crazy too. I've tried to keep the variable naming consistent at least by using `rosterEntry`. That's probably what the actual table/model should be changed to.

server/migration/1623276130329-FormatPhoneNumbers.ts
larskendall: There's no official phone column or type for Roster/RosterHistory, but it's probably a good idea to do something similar to the muster stuff with a sort of fuzzy search for one.

server/util/muster-utils.ts
larskendall: Looks like it can even be `Object.values()` and just skip the key altogether.

server/util/muster-utils.ts
larskendall: Good catch.

PR #156, opened 2021-06-15
Add support for dod id "in" query

pfrollo:
Handles "dirty" lists, e.g.  1000000001 ,,  ,1000000003 1000000004;1000000005

Screen Shot 2021-06-15 at 2 15 43 PM


src/components/query-builder/query-builder.tsx
```diff
@@ -414,6 +415,23 @@ export interface QueryBuilderProps {
...
 
+function toQuery(rows: QueryRow[]) {
+  const queryableRows: QueryRow[] = rows.filter(row => row.field && row.value !== '');
+  const query: QueryFilterState = {};
+  for (const { field, op, value } of queryableRows) {
+    if (value) {
+      if (op === 'in') {
+        // handle "dirty" lists, e.g.  1000000001 ,,  ,1000000003 1000000004;1000000005
+        const arrayOfValues = String(value).split(/[ ,;]+/).filter(vl => vl);
+        query[field!.name] = { op, value: arrayOfValues };
```
AnthonyWhitaker, line 426: I'm on my phone and can't verify the behavior at the moment... but the only thing that remotely concerns me is that even in the case of an invalid/empty array, this function will returns a valid query item for the row. To say it differently, it seems like the 'in' case allows an 'empty' value (an empty array) which seems to differ from the non "in" case which filters out the empty values (empty strings). Again, not sure if it's an actual issue just seems like potentially divergent behavior

pfrollo: Good catch, it is fixed now.

PR #158, opened 2021-06-21
update readme

pfrollo:
I updated README as I learned about the process of schema sync and migration.


README.md
```diff
@@ -103,12 +134,12 @@ npm test
 ./run-tests.sh
 ````
 
-#### Options
+### Options
```
richard3d, line 137: Do these apply to local-dev also? May want to add them if not.

richard3d (approved):
Had one small comment, but looks good to me otherwise.

README.md
pfrollo: Originally it was a subsection of Testing and I will make it such.

pfrollo: Never mind, it is a subsection....hmm...good question....

pfrollo: I don't know, I am leaving it as is.

PR #160, opened 2021-06-21
add Observation Type entity

pfrollo:
This is one of many incremental base feature branch PRs. See a more less a complete list of work needed here: https://app.asana.com/0/1198942245525292/1200474445920650/f   

This PR implements:

1. https://app.asana.com/0/1198942245525292/1200499903524476/f
2. https://app.asana.com/0/1198942245525292/1200499903524477/f
3. https://app.asana.com/0/1198942245525292/1200499903524482/f
4. https://app.asana.com/0/1198942245525292/1200499903524480/f
5. https://app.asana.com/0/1198942245525292/1200480586706040/f
6. https://app.asana.com/0/1198942245525292/1200480586706041/f
7. https://app.asana.com/0/1198942245525292/1200506195976675/f

With the _cascade_ options the controllers might not be needed: https://orkhan.gitbook.io/typeorm/docs/relations#cascade-options  However, I can see in the future we may want to restrict the observation types, for example when observations would be created in our app for example. In this case the observation types would be selected using a drop down list. In general, constraints are good. 

What would be the next step?

1. Secure observation controller (system to system)
2. Secure observation type service (system admin to system)
3. Observation.timestamp: we need to figure out what format the timestamp uses, see: https://wanago.io/2021/03/15/postgresql-typeorm-date-time/
4. Adjust column lengths bases on the source data lengths?
5. Discussion: What do we want to do when posting of some observations fail?
6. Do we need any observation POST payload validation before persisting? 

Screen Shot 2021-06-22 at 2 49 08 PM

Screen Shot 2021-06-22 at 9 31 45 AM


server/api/observation/observation.model.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  /* type of the observation symptoms
+  this is a FK to the observation_type table
+   */
+  @ManyToOne(() => ObservationType,
+    observationType => observationType.type,
+    { primary: false, onDelete: 'RESTRICT', eager: true, orphanedRowAction: 'nullify', cascade: false })
+  @JoinColumn({ name: 'type' })
+  @Column({ length: 50 })
+  type: string;
```
pfrollo, line 39: Replace with `type: ObservationType`

server/api/observation/type/observation-type-create.controller.ts
```diff
@@ -0,0 +1,17 @@
+import { Request, Response } from 'express';
+import { ObservationType } from './observation-type.model';
+
+class ObservationTypeCreateController {
+  async createObservationType(req: Request, res: Response) {
+    return res.json(await ObservationType.save(toObservationType(req.params.type)));
```
pfrollo, line 6: Instead of the query parameter used JSON body.

server/api/observation/observation-create.controller.ts
```diff
@@ -0,0 +1,25 @@
+import { Request, Response } from 'express';
+import { Observation } from './observation.model';
+import { ObservationType } from './type/observation-type.model';
+
+class ObservationCreateController {
+  async createObservation(req: Request, res: Response) {
```
AnthonyWhitaker, line 6: We tend to use more Typescript-esque conventions. These are the non-typed versions of Request/Response - we should probably follow similar conventions to the controllers and their action methods.

server/api/observation/observation-create.controller.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    console.log(JSON.stringify(req.body));
```
AnthonyWhitaker: Remove.

server/api/observation/observation-router.ts
```diff
@@ -0,0 +1,20 @@
...
+const observationRouter = express.Router() as any;
+
+/* Observation */
+observationRouter.get('/', getObservationController.getAllObservations);
+observationRouter.post('/', bodyParser.json(), createObservationController.createObservation);
+
+/* Observation Type */
+observationRouter.get('/type', getObservationTypeController.getAllObservationTypes);
+observationRouter.delete('/type/:type', deleteObservationTypeController.deleteObservationType);
+observationRouter.post('/type/:type', createObservationTypeController.createObservationType);
```
AnthonyWhitaker, line 18: You've obviously called this out already in your 'next steps' but we should probably get the security elements in before approving/merging this PR.

server/api/observation/observation.model.ts
```diff
@@ -0,0 +1,58 @@
...
+  We don't maintain the list of these IDs so there is no integrity constraint (foreign key).
+  */
+  @Column({ length: 10 })
+  edipi: string;
+
+  /* timestamp when the observation was reported
+  TODO: https://app.asana.com/0/1198942245525292/1200506195976678/f
+   */
+  @Column({ length: 50 })
+  timestamp: string;
```
AnthonyWhitaker, line 29: This timestamp would be very similar to the Orphaned Record timestamp (it's created from the same source DynamoDB data. Copying the declaration here for convenience.
```js
  @Column({
    type: 'timestamp',
    transformer: timestampColumnTransformer,
  })
  timestamp!: Date;
```

server/api/observation/observation.model.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  /* type of the observation symptoms
+  this is a FK to the observation_type table
+   */
+  @ManyToOne(() => ObservationType,
+    observationType => observationType.type,
+    { primary: false, onDelete: 'RESTRICT', eager: true, orphanedRowAction: 'nullify', cascade: false })
+  @JoinColumn({ name: 'type' })
+  @Column({ length: 50 })
+  type: ObservationType;
```
AnthonyWhitaker, line 39: Would this be better as a simple Enum type? I think it takes an actual code deployment to add new report/observation types in other parts of the system.

On the other hand, I wonder if this needs more thought in relation to the `ReportSchema` entity? Should it be FK'd there (Instead of the current or enum solutions?) Is a `ReportSchema` actually an `ObservationType`?

server/api/observation/observation.model.ts
```diff
@@ -0,0 +1,58 @@
...
+  // military unit description
+  @Column({ length: 100 })
+  unit: string;
+
+  /* military organization ID
+  Initially we will not make this a FK (foreign key) but it is to be considered.
+  If made a FK then we need to make sure we have all the FKs in the database and
+  also we would need to handle any resulting errors if a given FK is not found */
+  @Column({ length: 50 })
+  orgId: string;
```
AnthonyWhitaker, line 57: Should this be called `orgName`? If it is `orgId`, should it then be an optional number as in:  `orgId?: number`

server/api/observation/type/observation-type-create.controller.ts
AnthonyWhitaker: In theory, yes it would be a `body`, but going back to my previous comment... I wonder if we need ObservationType at all.

server/migration/1624397093677-AddObservation.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddObservation1624397093677 implements MigrationInterface {
+  name = 'AddObservation1624397093677';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`CREATE TABLE "observation_type" ("type" character varying(100) NOT NULL, CONSTRAINT "UQ_45daf0e906ab42b60abeaf83bd5" UNIQUE ("type"), CONSTRAINT "PK_45daf0e906ab42b60abeaf83bd5" PRIMARY KEY ("type"))`);
+    await queryRunner.query(`CREATE TABLE "observation" ("id" SERIAL NOT NULL, "document_id" character varying(100) NOT NULL, "edipi" character varying(10) NOT NULL, "timestamp" character varying(50) NOT NULL, "type" character varying(50) NOT NULL, "unit_id" character varying(50) NOT NULL, "unit" character varying(100) NOT NULL, "org_id" character varying(50) NOT NULL, CONSTRAINT "UQ_c3b96ba3699ceaa1c9c95bdab10" UNIQUE ("id", "document_id"), CONSTRAINT "PK_77a736edc631a400b788ce302cb" PRIMARY KEY ("id"))`);
+    await queryRunner.query(`ALTER TABLE "observation" ADD CONSTRAINT "FK_ea8a73e50440da7b4e5d5893d8b" FOREIGN KEY ("type") REFERENCES "observation_type"("type") ON DELETE RESTRICT ON UPDATE NO ACTION`);
+  }
```
AnthonyWhitaker, line 10: If we're sticking with ObservationTypes in a table... should we seed the current ObservationTypes here?

server/ormconfig.ts
```diff
@@ -42,10 +44,12 @@ export const ormConfig: PostgresConnectionOptions = {
     UserNotificationSetting,
     OrphanedRecord,
     OrphanedRecordAction,
+    Observation,
+    ObservationType,
   ],
   namingStrategy: new SnakeNamingStrategy(),
   synchronize: (env.isDev && process.env.SYNC_DATABASE === 'true'),
-  logging: false,
+  logging: (env.isDev || process.env.ORM_LOGGING === 'ture'),
```
AnthonyWhitaker, line 52: Remove. Or adjust to 1) fix the type in `true` and 2) maybe change `(env.isDev || ...` to `(env.isDev && ...` - I personally don't want the logging on just because I'm in development mode.

server/api/observation/observation-create.controller.ts
pfrollo: done

server/api/observation/observation-router.ts
pfrollo: Subtasks in Asana where created.

server/api/observation/observation.model.ts
pfrollo: Thanks! Done.

server/ormconfig.ts
pfrollo: Fixed

server/api/observation/observation-router.ts
pfrollo: Absolutely, I agree. This PR is for merging to my feature branch so it would be safe to merge all these changes before writing more code.

server/api/observation/observation.model.ts
pfrollo: Let's go with `orgName`. Do we want to omit `?` for now and confirm when we integrate whether this field is optional or not?

server/api/observation/type/observation-type-create.controller.ts
pfrollo: Thanks! This is also mentioned below so I am closing this.

server/api/observation/observation.model.ts
AnthonyWhitaker: Works for me.

server/api/observation/observation-create.controller.ts
pfrollo: Will be fixed as part of: https://app.asana.com/0/1198942245525292/1200516880894435/f

server/api/observation/observation.model.ts
pfrollo: Will use the report schema: https://app.asana.com/0/1198942245525292/1200516880894434/f

server/migration/1624397093677-AddObservation.ts
pfrollo: Will use report schema instead: https://app.asana.com/0/1198942245525292/1200516880894434/f

server/api/observation/type/observation-type.model.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseEntity, Entity, PrimaryColumn, Unique } from 'typeorm';
+
+@Entity()
+@Unique(['type'])
+export class ObservationType extends BaseEntity {
+
+  @PrimaryColumn({ length: 100 })
```
richard3d, line 7: This is declaring observation string as the primary key? Is this the string for the observation (like a description) itself or intended as an actual primary key?

pfrollo: ObservationType was removed. The latest WIP code is here: https://github.com/mystatus/covid19-reports/tree/add-observation-type-fk/server/api/observation

PR #162, opened 2021-06-24
Enhancement: Query Builder Expressions

richard3d:
Enhancement for https://app.asana.com/0/1188940305683068/1200506731385502/f.

Allows mathematic expressions for numeric values and time expressions via object input to momentJS `add()` method (see [here](https://momentjscom.readthedocs.io/en/latest/moment/03-manipulating/01-add/). Computed values are shown as preview when expressions are entered and invalid expressions are communicated to the user via the UI as well. Expressions are checked and evaluated client-side. The values are then sent server-side to perform the query. However, the expression itself and a boolean determining whether the query actually uses an expression, are actually sent to the server as well. Thus queries with expressions may be serialized and recomputed on the server in the future if desired.


richard3d:
Hmm expressions are behaving funny with `range` based queries, I will fix this before merging.

AnthonyWhitaker:
Nice! I was under the impression that server recomputed values were in scope for the current task. Do you think that's something we could get in before the end of the sprint?

richard3d:
@AnthonyWhitaker I was planning on saving that for serialization/deserialization of queries. That is the only place in code where it would need to be executed on the backend.

AnthonyWhitaker:
That serialization/deserialization already happens today to localStorage and should have already been finished as part of the last sprint (except I'm late in getting it done). The goal was to have them both finished simultaneously which I guess can now happen at the end of this sprint. Good thing I was late?!?

richard3d:
@AnthonyWhitaker, @pfrollo  one more review please?

AnthonyWhitaker (approved):

PR #164, opened 2021-06-25
Add observation

pfrollo:
https://app.asana.com/0/1198942245525292/1200474445920650/f


server/api/observation/observation-api-interface.ts
```diff
@@ -0,0 +1,9 @@
+export default interface ObservationApiInterface {
+  documentId: string;
+  edipi: string;
+  timestamp: Date;
+  unitId: string;
+  unit: string;
+  type_id: string;
```
pfrollo, line 7: rename to `typeId`

server/api/observation/observation-api-interface.ts
```diff
@@ -0,0 +1,9 @@
...
+  type_org: string;
```
pfrollo, line 8: rename to `typeOrgId`

server/migration/1624566131004-AddObservation.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddObservation1624566131004 implements MigrationInterface {
+  name = 'AddObservation1624566131004';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`CREATE TABLE "observation" ("id" SERIAL NOT NULL, "document_id" character varying(100) NOT NULL, "edipi" character varying(10) NOT NULL, "timestamp" TIMESTAMP NOT NULL, "unit_id" character varying(50) NOT NULL, "unit" character varying(100) NOT NULL, "type_id" character varying, "type_org" integer, CONSTRAINT "UQ_c3b96ba3699ceaa1c9c95bdab10" UNIQUE ("id", "document_id"), CONSTRAINT "PK_77a736edc631a400b788ce302cb" PRIMARY KEY ("id"))`);
```
pfrollo, line 7: Is it possible to annotate the model so that the default `type_org` becomes `org_id` or `type_org_id`?

server/api/observation/observation-api-interface.ts
pfrollo: done

server/api/observation/observation-api-interface.ts
pfrollo: done

server/migration/1624566131004-AddObservation.ts
larskendall: This is normally done by setting the `name` property on the `@Column` decorator's options. But I don't think `type_org` should even be showing up in the table like this. It probably has something to do with the erroneous "inverse side" setting in the `type` column.

server/api/observation/observation.router.ts
```diff
@@ -0,0 +1,13 @@
+import express from 'express';
```
larskendall, line 1: Maybe we should keep this `index.ts` to match the other routers for now, and rename them all in a single pass in a separate PR.

larskendall (changes requested):
This is exciting to be moving the observation data away from Elasticsearch!

There was one bug I think in the observation model which there's a note for. Pretty much everything else was mostly just style or organizational suggestions relating to consistency with the rest of the codebase.

server/migration/1624566131004-AddObservation.ts
pfrollo: These columns where generated by TypeORM, see report_schema:

`CREATE TABLE IF NOT EXISTS public.report_schema
(
    id character varying COLLATE pg_catalog."default" NOT NULL,
    name character varying(256) COLLATE pg_catalog."default" NOT NULL,
    columns json NOT NULL DEFAULT '[]'::json,
    org_id integer NOT NULL,
    CONSTRAINT "PK_c19c025244ceedaff94736f021f" PRIMARY KEY (id, org_id),
    CONSTRAINT "FK_d1ffa8699456464f8d833838cba" FOREIGN KEY (org_id)
        REFERENCES public.org (id) MATCH SIMPLE
        ON UPDATE NO ACTION
        ON DELETE CASCADE
)
`

server/api/observation/observation.api-i.ts
```diff
@@ -0,0 +1,9 @@
+export default interface ObservationApiI {
```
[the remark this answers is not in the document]
pfrollo: done

server/api/observation/observation.controller.create.ts
```diff
@@ -0,0 +1,27 @@
+import { getManager } from 'typeorm';
+import { Response } from 'express';
+import { ApiRequest, EdipiParam } from '../index';
+import ObservationApiI from './observation.api-i';
+import { timestampColumnTransformer } from '../../util/util';
+
+
+class ObservationControllerCreate {
```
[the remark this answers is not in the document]
pfrollo: Getting and saving entities are two different concerns and they should not be mixed up in one module. There should be a single module for a single concern.

server/api/observation/observation.model.ts
```diff
@@ -0,0 +1,46 @@
...
+  @ManyToOne(() => ReportSchema,
+    type => type.id,
+    { primary: false, onDelete: 'RESTRICT', eager: true, orphanedRowAction: 'nullify', cascade: false })
+  type: ReportSchema;
+
+  /* military unit description
+   Initially we will not make this a FK (foreign key) but it is to be considered.
+   If made a FK then we need to make sure we have all the FKs in the database and
+   also we would need to handle any resulting errors if a given FK is not found */
+  @Column({ length: 50 })
```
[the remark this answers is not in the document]
pfrollo: set to 300

server/api/observation/observation.model.ts
```diff
@@ -0,0 +1,46 @@
...
+  unitId: string;
+
+  // military unit description
+  @Column({ length: 100 })
```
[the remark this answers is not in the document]
pfrollo: set to 300

server/api/observation/observation.router.ts
```diff
@@ -0,0 +1,13 @@
...
+import getObservationController from './observation.controller.get';
+import createObservationController from './observation.controller.create';
+import { requireInternalUser } from '../../auth';
+
+
+const observationRouter = express.Router() as any;
+
+observationRouter.get('/', requireInternalUser, getObservationController.getAllObservations);
+observationRouter.post('/', requireInternalUser, express.json(), createObservationController.createObservation);
```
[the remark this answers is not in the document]
pfrollo: bodyParser.json() is deprecated so it should not be used. 
https://stackoverflow.com/questions/24330014/bodyparser-is-deprecated-express-4

server/api/observation/observation.router.ts
pfrollo: It does not make sense to follow bad habits and then have someone to undo them. The rest will be fixed as part of this ticket: https://app.asana.com/0/1188940305683068/1200549078853106

server/migration/1624566131004-AddObservation.ts
larskendall: It looks like `primary: true` is set on that column in `report-schema.model.ts`, so that's expected for the table to be generated like that.

The `id` on `ReportSchema` isn't automatically generated, so I suspect there's the possibility of id collisions, which is why `org_id` is part of the primary key. I can't remember the specific reason since I didn't write that `ReportSchema` model, but I recall JD saying this was necessary for some reason.

server/api/observation/observation.controller.create.ts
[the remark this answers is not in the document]
larskendall: In what way will this improve the codebase? I'd like to discuss it some more since it's a sweeping new convention you're proposing. These sorts of convention changes should be discussed before implementing to ensure that they bring value to the codebase, aren't just a stylistic preference, and consistency is maintained across the project.

This seems like a somewhat arbitrary distinction to me as far as code structure is concerned for this project, but I'm curious how you think this may help.

server/api/observation/observation.router.ts
larskendall: Why not just do it all at once as that separate task? I'd much rather keep everything in a consistent state.

If that tech debt task were to slip for some reason because higher priority tasks come up, then we would have multiple conventions in place. If this were to happen several times then the project will be a mess. Knowing how quickly this project has moved at times, as well as how often frequently people get moved on and off projects at PE, this is something we need to seriously consider when making these sorts of changes.

You're probably tired of me mentioning "consistency" at this point, but I think it's extremely important to maintain even when it may be a little uncomfortable. Making this file name inconsistent with the rest of the codebase right now isn't really saving you any time on that follow-up task anyways.

server/api/observation/observation.router.ts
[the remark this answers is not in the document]
larskendall: It's not technically deprecated according to the accepted answer to that question. It's just bundled into express now. If we're going to use `express.json()`, let's change it across the entire project all at once to avoid confusion. I'm fine with using either. I suppose eliminating the `body-parser` dependency would be nice.

server/migration/1624566131004-AddObservation.ts
pfrollo: Go it. As there is not action I am resolving this thread.

server/api/observation/observation.controller.create.ts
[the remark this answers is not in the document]
pfrollo: fixed

server/api/observation/observation.router.ts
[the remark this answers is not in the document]
pfrollo: I got a depreciation warning that is why I claimed it was deprecated.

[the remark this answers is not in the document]
pfrollo: fixed

server/api/observation/observation.router.ts
pfrollo: fixed

pfrollo:
Anthony wanted to move forward with the merge to unblock Richard.

PR #168, opened 2021-07-08
Update observation schema

pfrollo:
- Updated Observation based on FB from Anthony:
---- removed unit_id, type_org
---- added reporting_group
- Added Observation to ORM Config as it was a miss during a merge of the new project structure

### DIFF
--- a/server/src/api/observation/observation.controller.ts
+++ b/server/src/api/observation/observation.controller.ts
@@ -6,6 +6,7 @@ import { Log } from '../../util/log';
 import { ReportSchema } from '../report-schema/report-schema.model';
 import { RosterInfo, getRostersForIndividual } from '../roster/roster.controller';
 import { timestampColumnTransformer } from '../../util/util';
+import { BadRequestError } from '../../util/error-types';
 
 
 class ObservationController {
@@ -31,8 +32,7 @@ class ObservationController {
       }
     }
     Log.error(`Unable to save observation from: ${req.body}`);
-    return res.json(new Observation());
-
+    throw new BadRequestError('Unable to save observation');
   }
 }
 
--- a/server/src/api/observation/observation.controller.ts
+++ b/server/src/api/observation/observation.controller.ts
@@ -14,46 +14,53 @@ class ObservationController {
   }
 
   async createObservation(req: ApiRequest<EdipiParam, ObservationApiModel>, res: Response) {
-    return res.json(await createObservation(req));
-  }
-}
 
-function hasReportingGroup(roster: RosterInfo, reportingGroup: string | undefined) {
-  return reportingGroup && roster.unit.org?.reportingGroup === reportingGroup;
-}
+    Log.info('Creating observation', req.body);
 
-async function createObservation(req: ApiRequest<EdipiParam, ObservationApiModel>) {
-  Log.info('Creating observation', req.body);
+    const reportingGroup = req.body.reportingGroup;
+    const typeId = req.body.typeId;
 
-  const reportingGroup = req.body.reportingGroup;
-  const rosters: RosterInfo[] = await getRostersForIndividual((req.body.timestamp).toString(10), req.body.edipi);
+    const rosters: RosterInfo[] = await getRostersForIndividual(req.body.edipi, req.body.timestamp.toString(10));
 
-  for (const roster of rosters) {
-    if (hasReportingGroup(roster, reportingGroup)) {
-
-      const reportSchema = await ReportSchema.findOne({
-        relations: ['org'],
-        where: {
-          id: req.body.typeId,
-          org: roster.unit.org?.id,
-        },
-      });
-
-      if (reportSchema) {
-        const observation = new Observation();
-        observation.documentId = req.body.documentId;
-        observation.edipi = req.body.edipi;
-        observation.timestamp = timestampColumnTransformer.to(req.body.timestamp);
-        observation.type = reportSchema;
-        observation.unit = req.body.unit;
-        observation.reportingGroup = req.body.reportingGroup;
-        Log.info(`Saving new observation for ${observation.documentId} documentId`);
-        return observation.save();
+    for (const roster of rosters) {
+      if (hasReportingGroup(roster, reportingGroup)) {
+        const reportSchema = await findReportSchema(typeId, roster.unit.org?.id);
+        if (reportSchema) {
+          return res.json(saveObservationWithReportSchema(req, reportSchema));
+        }
       }
     }
+    Log.error(`Unable to save observation from: ${req.body}`);
+    return res.json(new Observation());
+
   }
-  Log.error(`Unable to save observation from: ${req.body}`);
-  return Promise.reject(new Observation());
+}
+
+function hasReportingGroup(rosterInfo: RosterInfo, reportingGroup: string | undefined) {
+  return reportingGroup && rosterInfo.unit.org?.reportingGroup === reportingGroup;
+}
+
+async function findReportSchema(reportSchemaId: string, orgId: number | undefined) {
+  const reportSchema = await ReportSchema.findOne({
+    relations: ['org'],
+    where: {
+      id: reportSchemaId,
+      org: orgId,
+    },
+  });
+  return reportSchema;
+}
+
+async function saveObservationWithReportSchema(req: ApiRequest<EdipiParam, ObservationApiModel>, reportSchema: ReportSchema) {
+  const observation = new Observation();
+  observation.documentId = req.body.documentId;
+  observation.edipi = req.body.edipi;
+  observation.timestamp = timestampColumnTransformer.to(req.body.timestamp);
+  observation.type = reportSchema;
+  observation.unit = req.body.unit;
+  observation.reportingGroup = req.body.reportingGroup;
+  Log.info(`Saving new observation for ${observation.documentId} documentId`);
+  return observation.save();
 }
 
 export default new ObservationController();
--- a/server/src/api/observation/observation.controller.ts
+++ b/server/src/api/observation/observation.controller.ts
@@ -1,9 +1,12 @@
 import { Request, Response } from 'express';
-import { getManager } from 'typeorm';
 import { ObservationApiModel } from './observation.types';
 import { Observation } from './observation.model';
 import { ApiRequest, EdipiParam } from '../api.router';
+import { Log } from '../../util/log';
+import { ReportSchema } from '../report-schema/report-schema.model';
+import { RosterInfo, getRostersForIndividual } from '../roster/roster.controller';
 import { timestampColumnTransformer } from '../../util/util';
+
 
 class ObservationController {
   async getAllObservations(req: Request, res: Response) {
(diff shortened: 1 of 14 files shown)


server/src/api/observation/observation.controller.ts
```diff
@@ -16,14 +16,15 @@ class ObservationController {
...
     .query(
-      'INSERT INTO public.observation(document_id, edipi, timestamp, unit_id, unit, type_id, type_org)'
-      + 'VALUES ($1, $2, $3, $4, $5, $6, $7)',
-      [req.body.documentId, req.body.edipi, timestampColumnTransformer.to(req.body.timestamp),
-        req.body.unitId, req.body.unit, req.body.typeId, req.body.orgId],
+      'INSERT INTO public.observation(document_id, edipi, timestamp, unit, type_id, reporting_group)'
+      + 'VALUES ($1, $2, $3, $4, $5, $6)',
+      [body.documentId, body.edipi, timeStamp, body.unit, body.typeId, body.reportingGroup],
     );
 }
```
larskendall, line 29: I think you can simplify this function down to this...

```typescript
async function createObservation(req: ApiRequest<EdipiParam, ObservationApiModel>) {
  Log.info('Creating observation', req.body);
  const observation = Observation.create(req.body);
  return observation.save();
}
```

NOTE: We're using `Log.info` in place of `console.log` throughout the codebase so that we have more control over when/how logs are displayed. For example, when using `Log`, all logs will be filtered out of test output by default.

server/src/api/observation/observation.model.ts
```diff
@@ -41,14 +40,10 @@ export class Observation extends BaseEntity {
...
-  @Column({ length: 300 })
-  unitId!: string;
-
-  // military unit description
+  // military unit
   @Column({ length: 300 })
   unit!: string;
+
+  @Column({ length: 100})
+  reportingGroup?: string;
```
larskendall, line 48: TypeORM doesn't pick up on the `?` when defining the schema, so you may have to do this to make sure the column is nullable (assuming that's the intention)...

```typescript
  @Column({
    length: 100,
    nullable: true,
  })
  reportingGroup?: string;
```

server/src/migration/1625762355386-ModifyObservation.ts
```diff
@@ -0,0 +1,18 @@
+import {MigrationInterface, QueryRunner} from 'typeorm';
```
larskendall, line 1: Apparently we don't have an eslint rule setup for this (I'll make a note to check that out), but the style we've been using is...

```typescript
import { SingleThing } from 'package';
import {
  FirstThing,
  SecondThing,
} from 'package';
```

larskendall (changes requested):
Looks good! Just a couple very minor suggested tweaks.

By the way, it's probably worth addressing this comment in this PR as well. I don't think it got fixed before merging the observation model stuff, and it may cause unexpected behavior at the moment.

https://github.com/mystatus/covid19-reports/pull/164#discussion_r661769354

server/src/migration/1625762355386-ModifyObservation.ts
pfrollo: Will be fixed by eslint.

server/src/api/observation/observation.controller.ts
pfrollo: Done, please review.

server/src/api/observation/observation.controller.ts
```diff
@@ -1,30 +1,65 @@
...
 
   async createObservation(req: ApiRequest<EdipiParam, ObservationApiModel>, res: Response) {
-    return res.json(await createObservation(req));
+
+    Log.info('Creating observation', req.body);
+
+    const reportingGroup = req.body.reportingGroup;
+    const typeId = req.body.typeId;
+
+    const rosters: RosterInfo[] = await getRostersForIndividual(req.body.edipi, req.body.timestamp.toString(10));
```
larskendall, line 23: I think this can be simplified a little.

```suggestion
    const rosterInfos = await getRosterInfosForIndividual(edipi, `${timestamp}`);
```

server/src/migration/1625857050805-ModifyObservation.ts
```diff
@@ -0,0 +1,18 @@
+import {MigrationInterface, QueryRunner} from 'typeorm';
```
larskendall, line 1: I need to see if there's an eslint rule for this...

```suggestion
import {
  MigrationInterface,
  QueryRunner,
} from 'typeorm';
```

server/src/api/observation/observation.model.ts
```diff
@@ -36,19 +35,13 @@ export class Observation extends BaseEntity {
...
-   also we would need to handle any resulting errors if a given FK is not found */
-  @Column({ length: 300 })
-  unitId!: string;
-
-  // military unit description
+  // military unit
   @Column({ length: 300 })
   unit!: string;
+
+  @Column({ length: 100, nullable: true})
```
larskendall, line 45: I need to figure out the eslint rule to point these out...

```suggestion
  @Column({
    length: 100,
    nullable: true,
  })
```

server/src/api/observation/observation.controller.ts
```diff
@@ -1,30 +1,65 @@
...
+      if (hasReportingGroup(roster, reportingGroup)) {
+        const reportSchema = await findReportSchema(typeId, roster.unit.org?.id);
+        if (reportSchema) {
+          return res.json(saveObservationWithReportSchema(req, reportSchema));
+        }
+      }
+    }
+    Log.error(`Unable to save observation from: ${req.body}`);
+    return res.json(new Observation());
+
```
larskendall, line 35:
```suggestion
```

server/src/api/observation/observation.controller.ts
```diff
@@ -1,30 +1,65 @@
...
-      [req.body.documentId, req.body.edipi, timestampColumnTransformer.to(req.body.timestamp),
-        req.body.unitId, req.body.unit, req.body.typeId, req.body.orgId],
-    );
+function hasReportingGroup(rosterInfo: RosterInfo, reportingGroup: string | undefined) {
+  return reportingGroup && rosterInfo.unit.org?.reportingGroup === reportingGroup;
+}
+
+async function findReportSchema(reportSchemaId: string, orgId: number | undefined) {
+  return ReportSchema.findOne({
+    relations: ['org'],
```
larskendall, line 45: I don't think you need this relation here. It doesn't appear to be used anywhere. I don't think it's necessary to query on the `org` property, but I could be wrong.

larskendall (changes requested):
Looks good overall. Just a few more suggestions.

server/src/api/observation/observation.controller.ts
```diff
@@ -14,46 +14,53 @@ class ObservationController {
...
+    for (const roster of rosters) {
+      if (hasReportingGroup(roster, reportingGroup)) {
+        const reportSchema = await findReportSchema(typeId, roster.unit.org?.id);
+        if (reportSchema) {
+          return res.json(saveObservationWithReportSchema(req, reportSchema));
+        }
       }
     }
+    Log.error(`Unable to save observation from: ${req.body}`);
+    return res.json(new Observation());
```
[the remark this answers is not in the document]
pfrollo: done

server/src/api/observation/observation.controller.ts
```diff
@@ -14,46 +14,53 @@ class ObservationController {
...
+  });
+  return reportSchema;
+}
+
+async function saveObservationWithReportSchema(req: ApiRequest<EdipiParam, ObservationApiModel>, reportSchema: ReportSchema) {
+  const observation = new Observation();
+  observation.documentId = req.body.documentId;
+  observation.edipi = req.body.edipi;
+  observation.timestamp = timestampColumnTransformer.to(req.body.timestamp);
+  observation.type = reportSchema;
```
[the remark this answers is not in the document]
pfrollo: I totally agree, it is confusing. It looks like ReportSchema is both, report schema and report type. Report type is the primary key. I would prefer that the ReportSchema is renamed to what it is. What is ReportSchema, how is is being used? I will make the change you suggested.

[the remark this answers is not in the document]
pfrollo: done

server/src/api/observation/observation.controller.ts
```diff
@@ -1,30 +1,65 @@
...
     return res.json(await Observation.find());
   }
 
   async createObservation(req: ApiRequest<EdipiParam, ObservationApiModel>, res: Response) {
-    return res.json(await createObservation(req));
+
+    Log.info('Creating observation', req.body);
+
+    const reportingGroup = req.body.reportingGroup;
+    const typeId = req.body.typeId;
```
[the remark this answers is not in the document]
pfrollo: done

server/src/api/observation/observation.controller.ts
```diff
@@ -1,30 +1,65 @@
...
+
 class ObservationController {
   async getAllObservations(req: Request, res: Response) {
     return res.json(await Observation.find());
   }
 
   async createObservation(req: ApiRequest<EdipiParam, ObservationApiModel>, res: Response) {
-    return res.json(await createObservation(req));
+
+    Log.info('Creating observation', req.body);
```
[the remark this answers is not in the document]
pfrollo: It looks like you removed an empty line? Done.

server/src/api/roster/roster.controller.ts
```diff
@@ -362,6 +328,45 @@ class RosterController {
 
 }
 
+export async function getRostersForIndividual(edipi: string, dateStr: string) {
```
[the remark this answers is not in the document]
pfrollo: There is already getRosterInfosForIndividual, please suggest a different name.

server/src/api/observation/observation.controller.ts
pfrollo: done

server/src/migration/1625857050805-ModifyObservation.ts
pfrollo: ok

pfrollo: IMO, this is a terrible style for the 21st century with large HD to 4k screens...why split it on multiple lines making us to scroll more? But whatever works for you....

server/src/api/observation/observation.model.ts
pfrollo: ok

server/src/api/observation/observation.controller.ts
pfrollo: Bad eslint setups?

server/src/api/observation/observation.controller.ts
pfrollo: I took this from an existing example. Each ORM is different. ORM knowledge is not directly transferable from one ORM to other given all the differences. ORM is convenient but makes us to guess the code, it is a good PR discussion started, it makes us dumber. :-)

pfrollo: Removal of `relations: ['org'],` did not work.

server/src/api/roster/roster.controller.ts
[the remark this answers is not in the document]
larskendall: Maybe `getRosterInfosForIndividualOnDate`?

server/src/migration/1625857050805-ModifyObservation.ts
larskendall: I personally find it a very worthwhile tradeoff to scroll more in order to keep things readable and to keep diffs extra clear, while also making it easy to trim/comment things out. I think it's much easier to scan code and catch mistakes/changes when you don't have to read it like a long sentence on one line. Even though we have such wide screens nowadays, I don't think it's a good idea to try and stretch code too far across them.

server/src/api/observation/observation.controller.ts
larskendall: Yeah I'm going to make a pass on our eslint rules here soon to account for these sorts of details.

server/src/api/observation/observation.controller.ts
larskendall: That's fine we can leave it.

Just out of curiosity, what happened when you removed it that stopped things from working?

server/src/api/roster/roster.controller.ts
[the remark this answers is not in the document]
pfrollo: done

PR #169, opened 2021-07-12
Start refactor of actions and reducers to use redux-toolkit

JWileczek:
This PR converts some of the existing actions and reducers to use redux-toolkit:
AppFrame
User
localStorageReducer
OrphanedRecords

Old actions and reducers have not yet been removed until functionality can/has been fully verified. I've done a rough pass over the functionality refactored here, but haven't fully tested every aspect of it yet. 

Converts separate files into single `slice` file. I.E. user.slice.ts which is an object that wraps actions and reducers. Reduces a large amount of boilerplate code surrounding async actions, and infers all action payloads nicely with typescript.

Can be used alongside traditional way of defining actions and reducers, so doesn't have to be a full swap to use this toolkit if it doesn't make sense. Existing Modal logic is a potential good case to keep standard since we do some custom behavior within that set that doesn't fit the async action model of toolkit.


client/src/reducers/local-storage.reducer.ts
```diff
@@ -1,48 +1,48 @@
...
-    }
-    case User.Actions.Logout.type: {
+    })
+    .addCase(UserActions.logout.type, () => {
       return localStorageInitialState;
-    }
-    case User.Actions.ChangeOrg.type: {
+    })
+    .addCase(UserActions.changeOrg.type, (state, action) => {
       const { orgId } = (action as User.Actions.ChangeOrg).payload;
```
larskendall, line 39: It looks like we can do this sort of thing to stay within the redux-toolkit style...

```suggestion
    .addCase(UserActions.changeOrg, (state, { payload }) => {
      const { orgId } = payload;
```

Apparently we never need to use the `type` property, since the action creators already override the `toString()` function.

client/src/reducers/local-storage.reducer.ts
```diff
@@ -1,48 +1,48 @@
...
   orgId?: number
   hideHelpCard?: Dict<boolean>
 }
 
 export const localStorageInitialState: LocalStorageState = {};
 
-export function localStorageReducer(state = localStorageInitialState, action: any): LocalStorageState {
-  switch (action.type) {
-    case User.Actions.Register.type: {
+export const localStorageReducer = createReducer(localStorageInitialState, builder => {
```
larskendall, line 18: Apparently we can still leverage `immer` when using the `createReducer` function. So each of these cases can probably be simplified a bit.

https://redux-toolkit.js.org/usage/usage-guide#simplifying-reducers-with-createreducer

client/src/slices/orphaned-record.slice.ts
```diff
@@ -0,0 +1,82 @@
...
+export const fetchCount = createAsyncThunk(
+  `${name}/fetchCount`,
+  async (orgId: number) => {
+    return OrphanedRecordClient.fetchCount(orgId);
+  },
+);
+
+export const fetchPage = createAsyncThunk(
+  `${name}/fetchPage`,
+  async (args: {orgId: number, page: number, limit: number, unit?: string}) => {
```
larskendall, line 34:
```suggestion
  async (args: { orgId: number, page: number, limit: number, unit?: string }) => {
```

client/src/slices/orphaned-record.slice.ts
```diff
@@ -0,0 +1,82 @@
...
+  },
+);
+
+export const orphanedSlice = createSlice({
+  name,
+  initialState: orphanedRecordInitialState,
+  reducers: {
+    clear: () => {
+      return { ...orphanedRecordInitialState};
+    },
```
larskendall, line 45: I suspect we can get away with just doing this, since redux-toolkit shouldn't be doing any internal mutations.

```suggestion
    clear: () => orphanedRecordInitialState,
```

client/src/slices/user.slice.ts
```diff
@@ -0,0 +1,184 @@
...
+}
+
+export interface DashboardPayload {
+  orgId: number
+  workspaceId: number
+  dashboardUuid: string
+}
+
+const name = 'user';
+const addFavoriteDashboard = createAsyncThunk(`${name}/addFavoriteDashboard`,
```
larskendall, line 41: Can we drop the first arguments onto a new line, consistent with `orphaned-record.slice.ts`?

client/src/slices/user.slice.ts
```diff
@@ -0,0 +1,184 @@
...
+      .addCase(removeFavoriteDashboard.rejected, (state, action) => {
+        const { workspaceId, dashboardUuid } = action.payload as DashboardPayload;
+        return addFavoriteDashboardImpl(state, workspaceId, dashboardUuid);
+      })
+      .addDefaultCase(state => state);
+
+  },
+});
+
+function addFavoriteDashboardImpl(state: UserState, workspaceId: number, dashboardUuid: string) {
```
larskendall, line 149: These `Impl` functions can probably be simplified pretty nicely by leveraging immer.

larskendall (changes requested):
This is great work! It's going to be amazing to get all of this stuff simplified so much.

Don't worry about all of the comments... most of them are either just simplifications or formatting suggestions.

Everything seemed to function ok, but I did get this exception repeating a bunch in the browser: `A non-serializable value was detected in an action...`

I suspect there must be some non-plain objects getting passed around into redux somehow. I'll let you dig into it some more.

PR #170, opened 2021-07-14
Add phone number to roster

pfrollo:
- prevent the server restarting frequently while writing code
- add ability to debug the server app
- make it easier to run locally by exporting INTERNAL_ACCESS_KEY
- add support for storing ph#s

### DIFF
--- a/server/nodemon.json
+++ b/server/nodemon.json
@@ -1,6 +1,7 @@
 {
   "ignore": ["**/*.test.ts", "**/*.spec.ts", ".git", "node_modules"],
   "watch": ["./"],
-  "exec": "ts-node ./src/app.ts",
-  "ext": "ts"
+  "exec": "node -r ts-node/register --inspect=127.0.0.1:9229 ./src/app.ts",
+  "ext": "ts",
+  "delay": 5000
 }
--- a/server/nodemon.json
+++ b/server/nodemon.json
@@ -2,6 +2,5 @@
   "ignore": ["**/*.test.ts", "**/*.spec.ts", ".git", "node_modules"],
   "watch": ["./"],
   "exec": "node -r ts-node/register --inspect=127.0.0.1:9229 ./src/app.ts",
-  "ext": "ts",
-  "delay": 5000
+  "ext": "ts"
 }
--- a/server/run-dev.sh
+++ b/server/run-dev.sh
@@ -2,6 +2,4 @@
 
 ./migration-run.sh
 
-export INTERNAL_ACCESS_KEY=anything
-
 npx nodemon
--- a/server/run-dev.sh
+++ b/server/run-dev.sh
@@ -2,4 +2,6 @@
 
 ./migration-run.sh
 
+export INTERNAL_ACCESS_KEY=anything
+
 npx nodemon
--- a/server/src/api/observation/observation.controller.ts
+++ b/server/src/api/observation/observation.controller.ts
@@ -4,10 +4,11 @@ import { Observation } from './observation.model';
 import { ApiRequest, EdipiParam } from '../api.router';
 import { Log } from '../../util/log';
 import { ReportSchema } from '../report-schema/report-schema.model';
-import rosterController, { RosterInfo, getRosterInfosForIndividualOnDate} from '../roster/roster.controller';
+import { RosterInfo, getRosterInfosForIndividualOnDate} from '../roster/roster.controller';
 import { timestampColumnTransformer } from '../../util/util';
 import { BadRequestError } from '../../util/error-types';
 import { assertRequestBody } from '../../util/api-utils';
+import { saveRosterPhoneNumber } from '../roster/roster-utils';
 
 
 class ObservationController {
@@ -31,7 +32,7 @@ class ObservationController {
       if (hasReportingGroup(roster, reportingGroup)) {
         const reportSchema = await findReportSchema(reportSchemaId, roster.unit.org?.id);
         if (reportSchema) {
-          await rosterController.savePhoneNumber(edipi, phoneNumber);
+          await saveRosterPhoneNumber(edipi, phoneNumber);
           return res.json(saveObservationWithReportSchema(req, reportSchema));
         }
       }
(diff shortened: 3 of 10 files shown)


server/nodemon.json
```diff
@@ -1,6 +1,7 @@
 {
   "ignore": ["**/*.test.ts", "**/*.spec.ts", ".git", "node_modules"],
   "watch": ["./"],
-  "exec": "ts-node ./src/app.ts",
-  "ext": "ts"
+  "exec": "node -r ts-node/register --inspect=127.0.0.1:9229 ./src/app.ts",
+  "ext": "ts",
+  "delay": 5000
```
larskendall, line 6: I don't like this delay since it slows down iteration speed. I'd rather we took it out.

What's the downside you've found with letting the server restart frequently? I haven't encountered any issues with it.

server/run-dev.sh
```diff
@@ -2,4 +2,6 @@
 
 ./migration-run.sh
 
+export INTERNAL_ACCESS_KEY=anything
```
larskendall, line 5: Will this overwrite a value set in `.env` at the project root? If so, I'd probably prefer to remove it to avoid confusion. I normally just have `INTERNAL_ACCESS_KEY` set in my `.env` file.

The current setup is a bit awkward, since using `server/run-dev.sh` won't load those environment variables but `run-dev-server.sh` will. Maybe we should just move the `.env` file to `server/.env` since I think we're only really messing with server values anyways. If we need to tweak anything on the client we could add one at `client/.env` eventually.

server/src/api/roster/roster.controller.ts
```diff
@@ -326,9 +326,17 @@ class RosterController {
     res.json(entry);
   }
 
+  async savePhoneNumber(edipi: string, phoneNumber: string) {
```
larskendall, line 329: Can we move this function out of the controller class? Our convention has been to only have endpoints in the controller itself. Typically, we'd put this sort of function in `roster-utils.ts` since it's used outside of this controller.

The function name may need to be made more explicit in that case, since the functions aren't wrapped in a class/namespace (which I'd like to fix at some point). So maybe something like `saveRosterPhoneNumber` would be good.

server/src/api/roster/roster.controller.ts
```diff
@@ -326,9 +326,17 @@ class RosterController {
...
+    const roster = await Roster.findOne({ where: { edipi }});
```
larskendall, line 330: The same `edipi` can potentially be in the roster table multiple times, since an individual can be on multiple units. So I think this should be a `Roster.find` call that updates potentially multiple roster entries.

server/src/api/observation/observation.controller.ts
```diff
@@ -18,18 +18,20 @@ class ObservationController {
...
 
-    const rosters: RosterInfo[] = await getRostersForIndividual(edipi, `${timestamp}`);
+    const rosters: RosterInfo[] = await getRosterInfosForIndividualOnDate(edipi, `${timestamp}`);
 
     for (const roster of rosters) {
       if (hasReportingGroup(roster, reportingGroup)) {
         const reportSchema = await findReportSchema(reportSchemaId, roster.unit.org?.id);
         if (reportSchema) {
+          await rosterController.savePhoneNumber(edipi, phoneNumber);
           return res.json(saveObservationWithReportSchema(req, reportSchema));
```
larskendall, line 35: It just occurred to me that this chunk should be wrapped in a transaction, since there are multiple database updates happening.

If you search "transaction" in the project you should see several examples of how we're doing those. You'll probably want to add a `manager` argument to the `savePhoneNumber` function as well and use that to save so that it stays within the transaction.

server/src/api/roster/roster-entity.ts
```diff
@@ -57,6 +57,11 @@ export abstract class RosterEntity extends BaseEntity {
   })
   lastName!: string;
 
+  @Column({
+    length: 20, default: 'Missing',
```
larskendall, line 61: What do you think about just making this column nullable, and letting null be the default? I'd rather leave the textual representation of a missing phone number up to the frontend, and making missing numbers null will let us check for non-existent phone numbers with `!phoneNumber`.

server/src/api/roster/roster-entity.ts
```diff
@@ -57,6 +57,11 @@ export abstract class RosterEntity extends BaseEntity {
...
+  })
+  phoneNumber!: string;
```
larskendall, line 63: We're currently formatting phone numbers with database hooks as they're updated, so that they're saved in a consistently format. You can see an example of that here...

https://github.com/mystatus/covid19-reports/blob/b615a8e9677aeca714d5d25eb2b00e5af81c9ced/server/src/api/access-request/access-request.model.ts#L77

larskendall (changes requested):
Just a few comments...

server/nodemon.json
pfrollo: The CPU fan goes out of control, it keeps restarting, reporting port in use, crashes, I have to restart it manually anyway but I will remove the setting. Overall, `nodemon` is not doing a good job at all.

server/run-dev.sh
pfrollo: fixed

server/src/api/roster/roster.controller.ts
pfrollo: fixed

server/src/api/roster/roster.controller.ts
pfrollo: Hmm....I did not know this!

server/src/api/roster/roster-entity.ts
pfrollo: done

server/src/api/roster/roster-entity.ts
pfrollo: Definitely a better!

pfrollo: done

server/src/api/roster/roster.controller.ts
pfrollo: done

server/src/api/observation/observation.controller.ts
pfrollo: done

PR #174, opened 2021-07-22
Refactor client requests and add shared project/workspace.

larskendall:
https://app.asana.com/0/1188940305683068/1200422935807053

The initial purpose of task was to refactor all of the direct `axios` calls that were scattered around and organize them into our client functions for hitting the api. However, in doing so there was a great opportunity to also setup shared types to really tighten up the api requests. I should have been more clever about doing this piecemeal, but unfortunately I didn't realize the extra complexity I was getting into and I let two huge refactors get intertwined. My apologies for such a ridiculously large PR.

Most of the changed files are just from moving `client` and `server` into a `packages` folder, which is the folder structure Lerna uses. We're not actually using Lerna (since Yarn Workspaces does everything we need) but I think this `packages` structure does help keep things more tidy. Getting the build working with this new setup was tricky, so it helped to get things as organized as possible.

Sometime soon we should tighten up the client and server even more by creating response types to ensure the api data models stay completely in sync between the two.

## Changes
- Created `packages` folder to house all of our workspaces.
- Created `shared` workspace to house any types or functions that are need on the client and server (eg. request query/body/response types).
- Split `client/src/client/index.ts` into multiple `xxxx.client.ts` files each corresponding to an api controllers.
- Moved all api calls using `axios` into `xxxx.client.ts` files.
- Renamed client-side api functions to match server controller function names.
- Updated scripts for new project structure.

## Notes
- Any request query/body/response types should be added to the `shared` workspace in the future. You'll just need to remember to export any new shared files in `shared/src/index.ts`. Maybe we can find a more hassle-free way of exporting those in the future, but this will work for now.


AnthonyWhitaker (approved):
Great stuff! Note that my approval is based on 10% code review and 90% functional testing. Thanks for doing all of this.

PR #178, opened 2021-07-27
Partial refactor to use redux-toolkit.

larskendall:
https://app.asana.com/0/1188940305683068/1200422935807035

This is the start of a larger refactor to switch all of the of the actions/reducers over to slices using redux-toolkit. John W had chosen these couple pieces to refactor, so I figured I'd see them through. I'll try to get the rest of the redux code switched over asap.

The redux-toolkit docs are worth skimming through btw if you're not familiar with it. Here are the sections that most directly relate to this PR...

https://redux-toolkit.js.org/usage/usage-with-typescript#createslice
https://redux-toolkit.js.org/usage/immer-reducers
https://redux-toolkit.js.org/usage/usage-with-typescript#createasyncthunk

There's some other stuff in there that might be really useful. I'm particularly interested in RTK Query, which is used to build out what we're doing in the client api functions, but in a way that's more closely integrated with redux.

## Changes
- Cleaned up store code for better typing.
- Removed plain object redux middleware and switched back to passing plain objects in actions. This was to clean up a ton of warnings that redux-toolkit was throwing when it detected non-serializable objects being passed into dispatch. Once the rest is refactored we wouldn't have needed that plain object middleware anyways.
- Added `useAppDispatch` and `useAppSelector` that are properly typed, and updated the project to use them everywhere instead of `useDispatch`/`useAppSelector`. They're actually taken directly from the docs [here](https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks)
- Converted `AppFrame`, `LocalStorage`, `OrphanedRecord`, and `User` over to use slices instead of actions/reducers.
- Added `assign` and `exportSlice` utility functions to make working with slices a little more clean/typesafe.


AnthonyWhitaker (approved):
Thanks for doing all of this. This is awesome!!

PR #180, opened 2021-07-29
Eslint updates

larskendall:
I figure while I'm making giant PRs I'll throw one more at you...

These updates should all be very low risk. The vast majority are automatic eslint fixes for the updated rules. There were just a few manual updates, mainly where I had to remove some unnecessary `async` keywords on function signatures where there was no `await` in the body.

The class whitespace padding rule is slightly annoying on all of the action classes, but those will all be going away anyways as we transition to redux-toolkit. So I'd say let's just put up with it for now.

## Updated Rules

- Require semicolons as interface/type delimiters - this is what TypeScript officially recommends (it's the default for the rule as well).
- Require promises to be awaited or marked `void` if the intention is not to handle them.
- Disallow `then` style of promise handling.
- Disallow callbacks in favor of promises.
- Disallow empty lines at the start/end of functions.
- Require empty lines at the start/end of classes.

## Other Updates

- Added `lint` script in `package.json` so that we can now run `yarn run lint` at the project root to lint every workspace at once.


AnthonyWhitaker (approved):
Works on my machine ™️ .

👍 💯 🥇

PR #183, opened 2021-07-30
Fix to maintain previous user-flow for seeding the postgres db

richard3d:


larskendall (approved):
Yay! Thanks for doing that. It works great.

PR #186, opened 2021-08-03
Added logic for spec and IT tests.

JWaldenPE:
1. Separate integration and unit tests:  https://app.asana.com/0/1188940305683068/1200662026396299/f


README.md
```diff
@@ -153,23 +153,31 @@ into a usable state.
...
 
 ### Options
-`--debug` - Prints all logs rather than only 'test' level logs.
 
-`--clean` - Drops, recreates, and migrates the test database before running tests.
+- `--debug` - Prints all logs rather than only 'test' level logs.
+
+- `--clean` - Server IT only.  Drops, recreates, and migrates the test database before running tests.
 
+- `--it | --spec` - Optional.  Limits the test to only run for IT or unit test.
```
pfrollo, line 168: I get this error: 

yarn test --spec 
yarn run v1.22.10
$ ./test.sh --spec
$ yarn workspace @covid19-reports/server test --spec
$ ./test.sh --spec
Warning: Cannot find any files matching pattern "**/*.spec.ts"

There is this "spec" test: packages/shared/src/utils/days.spec.ts

JWaldenPE: I can clarify in the documentation -- without specifying a workspace, test runs in server only by default.  Or, should the test iterate through each workspace?

pfrollo: 1. I see...Run it  with --spec, it gives the same error
2. I think `yarn test` should run all tests, not just the server tests

pfrollo:
mustering.oppoprtunities.it.ts is a unit test.

packages/shared/src/utils/days.spec.ts
```diff
@@ -33,6 +33,8 @@ describe('days', () => {
       expect(binaryDaysToDateArray(DaysOfTheWeek.Sunday | DaysOfTheWeek.Monday | DaysOfTheWeek.Tuesday | DaysOfTheWeek.Wednesday
       | DaysOfTheWeek.Thursday | DaysOfTheWeek.Friday | DaysOfTheWeek.Saturday)).to.eql([1, 2, 3, 4, 5, 6, 7]);
     });
+    it('expect to return next day Sunday to Monday', () => {
+      expect(binaryDaysToDateArray(nextDay(DaysOfTheWeek.Sunday))).to.eql([2]);
```
pfrollo, line 37: [2] can be replaced with [DaysOfTheWeek.Monday]

pfrollo: There are couple of more untested functions. I am not sure what the goal is but wonder if you plan to write tests for them.

packages/shared/utils.sh
```diff
@@ -0,0 +1,3 @@
+dir() {
```
pfrollo, line 1: Where is this being used?

packages/shared/src/utils/days.spec.ts
JWaldenPE: This was just working with the existing unit test you'd provided as an example-- I must have kept this change when I forked to work on the IT/unit tests.

packages/shared/utils.sh
JWaldenPE: Leftover from learning the test workflow-- I'll remove it.

README.md
JWaldenPE: Defaults to run all workspaces and can still use --it and --spec flags.  README updated.

packages/shared/src/utils/days.spec.ts
JWaldenPE: Updated

PR #187, opened 2021-08-03
Calculate and display muster compliance

pfrollo:
Extracted out test data generators for reuse.


packages/server/src/api/muster/muster.compliance.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { expect } from 'chai';
```
pfrollo, line 1: This is WIP...you can ignore it for now.

packages/server/src/api/muster/muster.compliance.ts
```diff
@@ -44,18 +44,14 @@ export function toMusterIntermediateCompliance(rosters: Roster[]): IntermediateM
 /**
```
pfrollo, line 44: Ignore, WIP

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -322,6 +322,7 @@ async function getUsersOnRosterByDate(orgId: number, unitName: string, date: str
     .select('DISTINCT ON(roster.edipi) roster.edipi, roster.change_type, unit.muster_configuration')
     .where(`roster.timestamp <= to_timestamp(:timestamp) AT TIME ZONE '+0'`, { timestamp })
     .andWhere('unit.name = :unitName', { unitName })
+    .andWhere('unit.org.id = :orgId', { orgId })
```
pfrollo, line 325: Fix for a bug we discovered with Richard.

packages/server/src/util/test-utils/roster.test-generator.ts
```diff
@@ -0,0 +1,39 @@
...
+import { getOrgTestData } from './org-test-generator';
+import { getAdminUserTestData } from './user.test-generator';
+import { getUnitsTestData } from './unit.test-generator';
+import { getReportSchemaTestData } from './report-schema.test-generator';
+
+export function getRosterTestData(numRosterEntries: number, customColumn: CustomRosterColumn, units: Unit[], numUnits: number) {
+  const rosterEntries: Roster[] = [];
+  for (let i = 0; i < numRosterEntries; i++) {
+    const customColumns: any = {};
+    customColumns[customColumn.name] = `custom column value`;
```
richard3d, line 15: Just a thought but we should change the the values here so better tests can be written for checking individual custom columns.

packages/server/src/util/test-utils/roster.test-generator.ts
```diff
@@ -0,0 +1,39 @@
...
+    const edipi = uniqueEdipi();
```
richard3d, line 16: uniqueEdipi is thorn in our side (as you already have pointed out) and even more annoying that we need to parse it as integer at times like below. Let's talk with Roel to see if this method can be adjusted.

packages/server/src/util/test-utils/seed.ts
```diff
@@ -146,53 +134,15 @@ export async function seedAll() {
...
-        { startTime: '2020-01-02T02:00:00.000', timezone: 'America/Los_Angeles', durationMinutes: 120, reportId: reportSchemas[1].id },
-      ],
-      includeDefaultConfig: true,
-    });
-    units.push(unit);
-  }
+  const units = getUnitsTestData(numUnits, org, reportSchemas);
   await Unit.save(units);
 
   const groupAdminRole = createGroupAdminRole(org);
```
richard3d, line 148: Should we have a corresponding generator for this (the role) also?

packages/server/src/util/test-utils/unit.test-generator.ts
```diff
@@ -0,0 +1,34 @@
+import { Org } from '../../api/org/org.model';
+import { ReportSchema } from '../../api/report-schema/report-schema.model';
+import { Unit } from '../../api/unit/unit.model';
+
+export function getUnitsTestData(numUnits: number, org: Org, reportSchemas: ReportSchema[]) {
```
richard3d, line 5: We should pull out the creation of musterConfigurations to a generator also (even though they are not serializable entities on their own), thoughts?

packages/server/src/util/test-utils/user.test-generator.ts
```diff
@@ -0,0 +1,14 @@
+import { User } from '../../api/user/user.model';
+import { uniqueEdipi, uniquePhone } from './unique';
+
+export function getAdminUserTestData() {
```
richard3d, line 4: We should probably parameterize this method more like the others.

richard3d (changes requested):
A few thoughts for the PR as a whole: We have these `generator` modules and then the methods are all named with `get`. I think we should keep naming consistent (like`generateUnit` or `generateUser` or even `createUnit` etc). Also the files are named redundantly (like instead of `test-generator` just name them `...-generator`. Also how do you feel about stashing these generator classes in a subdirectory called `data` in `test-utils` or something?

packages/server/src/api/muster/muster.controller.ts
pfrollo: removed

pfrollo:
> A few thoughts for the PR as a whole: We have these `generator` modules and then the methods are all named with `get`. I think we should keep naming consistent (like`generateUnit` or `generateUser` or even `createUnit` etc). Also the files are named redundantly (like instead of `test-generator` just name them `...-generator`. How do you feel about stashing these generator classes in a subdirectory called `data` in `test-utils` or something?

packages/server/src/util/test-utils/custom-roster-column.test-generator.ts
```diff
@@ -0,0 +1,15 @@
+import { RosterColumnType } from '@covid19-reports/shared';
+import { Org } from '../../api/org/org.model';
+import { CustomRosterColumn } from '../../api/roster/custom-roster-column.model';
+
+export function getCustomRosterColumnTestData(org: Org, customColumnOrgCount: number) {
```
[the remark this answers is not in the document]
pfrollo: This is copy/paste from seed.ts. This request is out of the scope of this PR I think.

packages/server/src/util/test-utils/org-test-generator.ts
```diff
@@ -0,0 +1,14 @@
+import { User } from '../../api/user/user.model';
+import { Org } from '../../api/org/org.model';
+
+export function getOrgTestData(admin: User, orgCount: number) {
+  const org = Org.create({
```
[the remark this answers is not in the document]
pfrollo: This is copy/paste from seed.ts. This request is out of the scope of this PR I think.

packages/server/src/util/test-utils/report-schema.test-generator.ts
```diff
@@ -0,0 +1,15 @@
+import { Org } from '../../api/org/org.model';
+import { defaultReportSchemas, ReportSchema } from '../../api/report-schema/report-schema.model';
+
+export function getReportSchemaTestData(org: Org) {
+  let reportSchemas = ReportSchema.create(defaultReportSchemas);
+  reportSchemas = reportSchemas.concat(ReportSchema.create(defaultReportSchemas));
```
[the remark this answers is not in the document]
pfrollo: This is copy/paste from seed.ts. This request is out of the scope of this PR I think.

packages/server/src/util/test-utils/roster.test-generator.ts
pfrollo: This is copy/paste from seed.ts. This request is out of the scope of this PR I think.

packages/server/src/util/test-utils/roster.test-generator.ts
pfrollo: I think it is fine as is,  we have a bigger fish to fry. Optionally, we can remove the parse Int call and append the edipi as is or generate better names. This is out of scope I think for now.

packages/server/src/util/test-utils/seed.ts
pfrollo: Yes, and there is lot more that should be extracted. I concentrated on what is needed to get roster and all the related data generated. 
I think this work needs to be part of a new task so that I can move back to completing my primary assignment. 
This is a good discussion and some initial work done with more needed.

packages/server/src/util/test-utils/unit.test-generator.ts
pfrollo: Yes, and there is lot more that should be extracted. I concentrated on what is needed to get roster and all the related data generated. 
I think this work needs to be part of a new task so that I can move back to completing my primary assignment. 
This is a good discussion and some initial work done with more needed.

packages/server/src/util/test-utils/user.test-generator.ts
pfrollo: Yes, but out of the scope for now.

PR #189, opened 2021-08-06
Unit tests shared

JWaldenPE:
Run a postgresql container on 5432, only if there isn't one already:
https://github.com/mystatus/covid19-reports/blob/47fd3c9f3f2a60850dc48418d92d52cee0f48c60/packages/server/test.sh#L24-L29

### DIFF
--- a/packages/server/test.sh
+++ b/packages/server/test.sh
@@ -21,6 +21,13 @@ while [[ "$#" -gt 0 ]]; do
   shift
 done
 
+docker rm -f unittestpg || true
+docker run -itd --name unittestpg -p 5432:5432 -e POSTGRES_PASSWORD=postgres postgres
+
+while ! pg_isready -h localhost -p 5432; do
+  sleep 1
+done
+
 export NODE_ENV="test"
 export SYNC_DATABASE="false"
 export SQL_DATABASE="dds_test"
@@ -50,3 +57,5 @@ npx ts-mocha \
   --config ".mocharc.yml" \
   --exit \
   $TYPE
+
+docker rm -f unittestpg || true
--- a/packages/server/test.sh
+++ b/packages/server/test.sh
@@ -21,12 +21,12 @@ while [[ "$#" -gt 0 ]]; do
   shift
 done
 
-docker rm -f unittestpg || true
-docker run -itd --name unittestpg -p 5432:5432 -e POSTGRES_PASSWORD=postgres postgres
-
-while ! pg_isready -h localhost -p 5432; do
-  sleep 1
-done
+if ! ph_isready -h localhost -p 5432; then
+  docker run -itd --name unittestpg -p 5432:5432 -e POSTGRES_PASSWORD=postgres postgres
+  while ! pg_isready -h localhost -p 5432; do
+    sleep 1
+  done
+fi
 
 export NODE_ENV="test"
 export SYNC_DATABASE="false"
--- a/packages/shared/src/utils/days.spec.ts
+++ b/packages/shared/src/utils/days.spec.ts
@@ -1,7 +1,38 @@
 import {expect} from 'chai';
-import { binaryDaysToDateArray, DaysOfTheWeek, nextDay, dayIsIn } from './days';
+import { binaryDaysToDateArray, DaysOfTheWeek, nextDay, daysToString, dayIsIn } from './days';
 
 describe('days', () => {
+  describe('daysToString()', () => {
+    it('expect to return Sun', () => {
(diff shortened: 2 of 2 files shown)


packages/server/test.sh
```diff
@@ -21,6 +21,13 @@ while [[ "$#" -gt 0 ]]; do
   shift
 done
 
+if ! ph_isready -h localhost -p 5432; then
```
pfrollo, line 24: What does `ph_isready` do? Is this something I need to install?

pfrollo: Oh...it is a typo I think...pg_isready?

pfrollo: How does the script behave when pg_isready is not installed? Worth testing it?

pfrollo, line 24: Would it be worthwhile, ask Roel, to move test.sh to a **scripts** folder and then extract out this function into a separate specialized script responsible only for starting the DB also stored in the **scripts** folder?

pfrollo: Then also there would be a separate script for stopping the DB. See the last line: **docker rm -f unittestpg || true**

packages/shared/src/utils/days.spec.ts
```diff
@@ -33,6 +64,13 @@ describe('days', () => {
       expect(binaryDaysToDateArray(DaysOfTheWeek.Sunday | DaysOfTheWeek.Monday | DaysOfTheWeek.Tuesday | DaysOfTheWeek.Wednesday
       | DaysOfTheWeek.Thursday | DaysOfTheWeek.Friday | DaysOfTheWeek.Saturday)).to.eql([1, 2, 3, 4, 5, 6, 7]);
     });
+  });
+  describe('dayIsIn()', () => {
+    it('expect to return true', () => {
```
pfrollo, line 69: I can't tell by reading the test description what is being tested.

packages/shared/src/utils/days.spec.ts
```diff
@@ -33,6 +64,13 @@ describe('days', () => {
...
+      expect(dayIsIn(DaysOfTheWeek.Friday, DaysOfTheWeek.Sunday | DaysOfTheWeek.Monday | DaysOfTheWeek.Friday)).to.eql(true)
+    });
+  });
+  describe('nextDay()', () => {
```
pfrollo, line 73: I can't tell by reading the test description what is being tested.

packages/shared/src/utils/days.spec.ts
pfrollo: I know coming up with good test descriptions is not always easy.

pfrollo: How about: "Saturday should be part of a weekend" and then modify slightly the test to test this scenario?

packages/server/test.sh
JWaldenPE: Fixed the typo-- pg_isready is a utility that comes along with Postgres to make sure a db is ready.  It wasn't enough to check for the listening port.  I'll add a check for the pg_isready utility.

packages/server/utils.sh
```diff
@@ -1,5 +1,18 @@
...
 
+database_docker() {
+  if [ "$1" = "start" ]; then
+    if ! pg_isready -h localhost -p 5432; then
+      docker run -itd --name unittestpg -p 5432:5432 -e POSTGRES_PASSWORD=postgres postgres
+      while ! pg_isready -h localhost -p 5432; do
+        sleep 1
+      done
+    fi
+  elif [ "$1" = "stop" ]; then
```
pfrollo, line 11: I would `echo`  "Stopping docker DB...."

packages/server/test.sh
pfrollo: Good catch! This is good. Thanks!

packages/server/utils.sh
JWaldenPE: https://github.com/mystatus/covid19-reports/blob/bfe2a2ca5223a9a0fc71eb069d1ef9fa2c0c9f90/packages/server/utils.sh#L23-L25

packages/server/utils.sh
```diff
@@ -1,5 +1,18 @@
 #!/bin/bash
 
+database_docker() {
+  if [ "$1" = "start" ]; then
```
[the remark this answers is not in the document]
JWaldenPE: https://github.com/mystatus/covid19-reports/blob/bfe2a2ca5223a9a0fc71eb069d1ef9fa2c0c9f90/packages/server/utils.sh#L4-L10

PR #191, opened 2021-08-06
Add unit muster compliance table. Update back-end. to support getting …

richard3d:
This PR supports the rendering/plotting of compliance data per unit on the muster page, based on date-range queries. Backend support was added to support date ranges in single query/request.

### DIFF
--- a/packages/client/src/client/muster.client.ts
+++ b/packages/client/src/client/muster.client.ts
@@ -1,4 +1,8 @@
 import {
+  GetMusterComplianceByDateQuery,
+  GetMusterComplianceByDateRangeQuery,
+  GetMusterComplianceByDateResponse,
+  GetMusterComplianceByDateRangeResponse,
   GetMusterRosterQuery,
   GetMusterUnitTrendsQuery,
 } from '@covid19-reports/shared';
@@ -20,6 +24,18 @@ export class MusterClient {
 
   static getMusterUnitTrends(orgId: number, query: GetMusterUnitTrendsQuery): Promise<ApiMusterTrends> {
     return client.get(`${orgId}/unit-trends`, {
+      params: query,
+    });
+  }
+
+  static getMusterComplianceByDate(orgId: number, unitName: string, query: GetMusterComplianceByDateQuery): Promise<GetMusterComplianceByDateResponse> {
+    return client.get(`${orgId}/${unitName}/complianceByDate`, {
+      params: query,
+    });
+  }
+
+  static getMusterComplianceByDateRange(orgId: number, unitName: string, query: GetMusterComplianceByDateRangeQuery): Promise<GetMusterComplianceByDateRangeResponse> {
+    return client.get(`${orgId}/${unitName}/complianceByDateRange`, {
       params: query,
     });
   }
--- a/packages/client/src/components/pages/muster-page/muster-page.tsx
+++ b/packages/client/src/components/pages/muster-page/muster-page.tsx
@@ -1,15 +1,17 @@
 import MomentUtils from '@date-io/moment';
 import {
   Box,
-  Card,
-  CardContent,
   Container,
   Grid,
   MenuItem,
   Paper,
   Select,
+  Table,
+  TableBody,
   TableContainer,
-  Typography,
+  TableCell,
+  TableHead,
+  TableRow,
 } from '@material-ui/core';
 import {
   DateTimePicker,
@@ -42,9 +44,7 @@ import { MusterPageHelp } from './muster-page-help';
 import useStyles from './muster-page.styles';
 import {
   ApiMusterRosterEntriesPaginated,
-  ApiMusterTrends,
   ApiRosterColumnInfo,
-  ApiUnitStatsByDate,
 } from '../../../models/api-response';
 import { ButtonWithSpinner } from '../../buttons/button-with-spinner';
 import { UnitSelector } from '../../../selectors/unit.selector';
@@ -68,14 +68,12 @@ export const MusterPage = () => {
   const user = useAppSelector(state => state.user);
 
   const maxNumColumnsToShow = 6;
-  const maxTopUnitsCount = 5;
   const trendChart = {
     layout: {
-      height: 320,
+      height: 128,
       autosize: true,
-      barmode: 'stack',
+      barmode: 'group',
       yaxis: {
-        title: 'Non-Muster Rate %',
         color: '#A9AEB1',
         hoverformat: '.0f%',
         showticklabels: false,
@@ -99,13 +97,10 @@ export const MusterPage = () => {
     } as Partial<Plotly.Config>,
   };
 
-  const weeklyTrendChart = {
+  const complianceBarChart = {
     ...trendChart,
   };
 
-  const monthlyTrendChart = {
-    ...trendChart,
-  };
 
   const [rosterSelectedTimeRangeIndex, setRosterSelectedTimeRangeIndex] = usePersistedState('musterRosterSelectedTimeRangeIndex', 0);
   const [rosterFromDateIso, setRosterFromDateIso] = usePersistedState('musterRosterFromDateIso', moment().startOf('day').toISOString());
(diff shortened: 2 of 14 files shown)


richard3d:
Rebase completed and WIP stuff from @pfrollo removed or TODO'd to ensure builds work still.

packages/client/src/client/muster.client.ts
```diff
@@ -24,4 +28,16 @@ export class MusterClient {
     });
   }
 
+  static getMusterComplianceByDate(orgId: number, unitName: string, query: GetMusterComplianceByDateQuery): Promise<GetMusterComplianceByDateResponse> {
```
pfrollo, line 31: This function appears not to be used.

packages/client/src/client/muster.client.ts
```diff
@@ -24,4 +28,16 @@ export class MusterClient {
...
+    return client.get(`${orgId}/${unitName}/complianceByDate`, {
+      params: query,
+    });
+  }
+
+  static getMusterComplianceByDateRange(orgId: number, unitName: string, query: GetMusterComplianceByDateRangeQuery): Promise<GetMusterComplianceByDateRangeResponse> {
```
pfrollo, line 37: By looking at the **GetMusterComplianceByDateRangeQuery** type I can't tell how to construct the **query** parameter.  Since this is a range, would it be more clear to pass in an object with a start and end date instead of a single string? 
{ isoStartDate: string; isoEndDate: string }

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -231,30 +233,76 @@ class MusterController {
...
    */
   async getMusterComplianceByDate(
     req: ApiRequest<{ orgId: number; unitName: string }, null, GetMusterComplianceByDateQuery>,
     res: Response<GetMusterComplianceByDateResponse>,
   ) {
-    const outValue = { musterComplianceRate: 0 };
-    const reportDate = dateFromString(req.query.timestamp);
-    if (!reportDate) {
-      throw new BadRequestError('Missing reportDate.');
+    const outValue = { musterComplianceRate: 0, isoDate: req.query.isoDate };
```
pfrollo, line 242: This approach below gets the value 1st and then sets the response, this flow seems more logical to me...This might be a matter of a taste. 
```suggestion
      const musterComplianceRate = await getMusterComplianceByDate(req.params.orgId, req.params.unitName, req.query.isoDate);
    res.json({ musterComplianceRate, isoDate: req.query.isoDate });
```

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -231,30 +233,76 @@ class MusterController {
 
   /**
    * This method returns the normalized rate (0 - 1.0) of compliance on a given
-   * date for a given unit, against  of the unit's muster configs.
+   * date for a given unit, against all of the unit's muster configs.
    */
   async getMusterComplianceByDate(
```
pfrollo, line 238: This module has 400 LOC already and I wonder if there is a way to split it to smaller modules, if you feel like splitting it.

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -231,30 +233,76 @@ class MusterController {
...
+}
+
+/**
+ * This helper method returns the normalized rate (0 - 1.0) of compliance on a given
+ * date for a given unit, against all of the unit's muster configs.
+ * @param orgId - the organization id
+ * @param unitName - the roster the users belong to
+ * @param isoDate - the date in format YYYY-MM-DD
+ */
+async function getMusterComplianceByDate(orgId: number, unitName: string, isoDate: string) {
```
pfrollo, line 283: This function is fairly long making it harder to read given the size (every single line needs to be read). When I see comments like this...
// handle one-time muster
// handle recurring muster...
...typically it indicates areas of code which can be extracted out as functions.

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -231,30 +233,76 @@ class MusterController {
...
-    const usersOnRoster = await getUsersOnRosterByDate(req.params.orgId, req.params.unitName, req.query.timestamp);
+    const usersOnRoster = await getUsersOnRosterByDate(orgId, unitName, isoDate);
     const users = usersOnRoster.users;
     const userCount = users.length;
     let musterAvg: number = 0;
 
     if (usersOnRoster.musterConfig.length === 0 || userCount === 0) {
-      res.json(outValue);
-      return;
+      return outValue;
```
pfrollo, line 302: You could simplify it by returning zero here and when **reportDate** is **falsy** (fail fast). By declaring **outValue = 0** at the top and using it in the middle and very bottom of the function makes one look for what the value is at these locations which is not necessary.

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -231,30 +233,76 @@ class MusterController {
...
+  let reportDate: Date | undefined;
+  try {
+    reportDate = dateFromString(isoDate);
+  } catch (DateParseError) {
+    throw new BadRequestError('Invalid ISO date.');
+  }
+
+  if (reportDate) {
     // eslint-disable-next-line no-bitwise
     const dayBit = 1 << reportDate.getDay();
```
pfrollo, line 294: The days.ts performs similar operations. Would it be a good idea to wrap this into a method inside the days.ts module?

packages/client/src/components/pages/muster-page/muster-page.tsx
```diff
@@ -448,6 +419,34 @@ export const MusterPage = () => {
     setRosterToDateIso(range.toDateIso);
   };
 
+  const renderUnitComplianceTableRows = () => {
```
pfrollo, line 422: I would prefer this React component to be stored in its own file and **props** used to pass down the data. It might be too much work do actually do this.

packages/client/src/components/pages/muster-page/muster-page.tsx
```diff
@@ -599,69 +598,28 @@ export const MusterPage = () => {
...
-                <Plot
-                  style={{ width: '100%' }}
-                  data={monthlyTrendData}
-                  layout={monthlyTrendChart.layout}
-                  config={monthlyTrendChart.config}
-                />
-              </CardContent>
-            </Card>
-          </Grid>
+          {/* Muster compliance Table */}
```
pfrollo, line 601: It looks like this is a React component which could be extracted out into a separate file.

packages/client/src/client/muster.client.ts
richard3d: I tried to indicate by the name that it is an `isoDateRange` which is a standardized format `YYYY-MM-DDT00:00:00/YYYY-MM-DDT00:00:00`, but perhaps I should use a start and end as part of the query params to be more explicit as opposed to body params? I think this pattern is used for other GET endpoints that want a date range.

packages/client/src/components/pages/muster-page/muster-page.tsx
```diff
@@ -224,25 +217,53 @@ export const MusterPage = () => {
...
-        weeksCount: '6',
-        monthsCount: '6',
-      });
-      setRawTrendData(data);
+      // determine if we are displaying for all units or just a single unit
+      const unitsToDisplay = rosterUnitId === -1 ? units : [units.find(unit => unit.id === rosterUnitId)];
+      const plotData = new Map();
+      const overallComplianceData = new Map();
+      const todaysComplianceData = new Map();
+      for (let i = 0; i < unitsToDisplay.length; i++) {
```
pfrollo, line 227: Can be replaced with **forEach** and not to worry about the index "i".

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -231,30 +233,76 @@ class MusterController {
...
     let musterAvg: number = 0;
 
     if (usersOnRoster.musterConfig.length === 0 || userCount === 0) {
-      res.json(outValue);
-      return;
+      return outValue;
     }
 
+    let numActiveConfigs: number = 0;
     for (let i = 0; i < usersOnRoster.musterConfig.length; i++) {
```
pfrollo, line 306: Can be replaced with **forEach** and not to worry about the index "i" I think.

pfrollo (approved):

packages/client/src/client/muster.client.ts
pfrollo: I personally feel that joining two pieces of data into a single string, even if related, is not the best idea. The names you used are great, no issue there.

packages/server/src/api/muster/muster.controller.ts
richard3d: This is valid concern, I think one thing that is a little odd with patterns that were used for the API originally, is we do not separate "business logic" into services (i.e. models, services, controllers). This might help a little bit at least.

pfrollo: The list of imports is out of control. While this is a React "single page" component, quickly looking into the imports, I don't see imports of sub-components other than very low level ones. I would expect bunch of higher level components imported.

packages/client/src/components/pages/muster-page/muster-page.tsx
richard3d: This is probably part of my own paranoia but in general when looping and calling async functions inside the loop I do NOT use higher order functions like this. see: https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

packages/server/src/api/muster/muster.controller.ts
richard3d: I think this comment is misplaced...you mean the page rather than the controller?

packages/server/src/api/muster/muster.controller.ts
richard3d: Actually forgot this loop contains async calls too :/

packages/server/src/api/muster/muster.controller.ts
pfrollo: Yes, the last one was meant for the impacted React component.

packages/server/src/api/muster/muster.controller.ts
pfrollo: The **if** and **else** blocks represent about 2/3 of the code in this method. Both code blocks can be extracted out to functions. When extracted, it helps with keeping the scope of local variables small, the scope not extending to all parts of the large function.

packages/client/src/client/muster.client.ts
richard3d: resolved with query params now

packages/client/src/components/pages/muster-page/muster-page.tsx
richard3d: I will address the component refactors in separate PR

packages/client/src/components/pages/muster-page/muster-page.tsx
richard3d: See comment above

packages/client/src/client/muster.client.ts
richard3d: I left this as a convenience method (so range didn't have to be used) but we can add to the chopping block in another PR

PR #192, opened 2021-08-09
fix: make timestamp w/o TZ to be interpreted as UTC

pfrollo:
https://app.asana.com/0/1188940305683068/1200740857713872/f


richard3d (approved):

richard3d:
@pfrollo I approve, but let's make sure we get sign off by Lars or someone on the project longer just to make sure there are no other implications.

larskendall (approved):
This looks good to me. As long as tests are passing I think we should be fine. I'd be surprised if this caused any issues, since it's the behavior we would have expected in the first place.

Great find on that solution btw!

PR #193, opened 2021-08-10
Add muster calculation

pfrollo:


packages/server/src/api/muster/muster.compliance.spec.ts
```diff
@@ -1,19 +1,79 @@
...
+        });
+        it(`should return compliance percentage`, () => {
+          const musterCompliance: MusterCompliance[] = calculateMusterCompliance(observations, musteringOpportunities, rosterForMusterCompliance);
+          expect(musterCompliance[0].musterPercent).equal(100);
+        });
+        it(`should return compliance`, () => {
+          const musterCompliance: MusterCompliance[] = calculateMusterCompliance(observations, musteringOpportunities, rosterForMusterCompliance);
+          expect(musterCompliance[0].mustersReported).equal(1);
+        });
+        it(`should return 100% compliance when mustering is not required`, () => {
```
richard3d, line 72: I suggest returning null when compliance is not required so the UI eventually has some way to know (it can still just render 100 percent on the graph). Plus calculations like average compliance (if desired) will be thrown off too.

packages/server/src/api/muster/muster.compliance.ts
```diff
@@ -45,29 +46,71 @@ export function toMusterIntermediateCompliance(rosters: Roster[]): IntermediateM
...
   mustersReported: number;
   musterPercent: number;
-} & IntermediateMusterCompliance;
+} & RosterForMusterCompliance;
 
 /**
  * Represents muster compliance without any compliance calculations
  */
-type IntermediateMusterCompliance = {
+export type RosterForMusterCompliance = {
```
richard3d, line 113: This type seems like it could use a slightly more indicative name like `RosterEntry` `UserInfo` or `UserOnRoster` etc

packages/server/src/api/muster/muster.compliance.ts
```diff
@@ -45,29 +46,71 @@ export function toMusterIntermediateCompliance(rosters: Roster[]): IntermediateM
...
+
+    if (!unitMusteringOpportunities || unitMusteringOpportunities.length === 0) {
+      return { totalMusters: 0, mustersReported: 0, musterPercent: 100, ...roster };
+    }
+
+    const totalMusters = unitMusteringOpportunities.length;
+
+    const edipiObservations = filterObservationsByEdipi(observations, edipi);
+    const mustersReported = calculateIndividualMusterCompliance(edipiObservations, unitMusteringOpportunities);
+    const musterPercent = (100 * mustersReported) / totalMusters;
```
richard3d, line 65: order of ops doesn't matter with multiplication and division ;), generally like to see the mult come second when normalizing values but may just be me lol

packages/server/src/api/muster/muster.postgres.controller.ts
```diff
@@ -38,7 +38,6 @@ class MusterPostgresCtr {
       const rosters = await getRosterWithUnitsAndEdipis(unitId, orgId);
```
richard3d, line 38: I don't think this is doing what you ultimately want it to do :/...yes you are finding folks "on roster" but these are only recent (i.e. the current roster they are on) and you have no idea if the user was on that roster for a given date. In order to know that you must query the `roster_history` table. You need that so you can make sure that they were on the roster for a given unit on a given date in order to determine if an observation is actually expected from them first.

packages/server/src/api/muster/muster.compliance.ts
```diff
@@ -45,29 +46,71 @@ export function toMusterIntermediateCompliance(rosters: Roster[]): IntermediateM
...
+  return compliance;
 }
 
+function filterObservationsByEdipi(observations: RecordedObservations[], edipi: string): RecordedObservationTimestamp[] {
+  return observations
+    .filter(observation => observation.edipi === edipi)
+    .map(observation => { return { timestamp: observation.timestamp }; });
+}
+
+function calculateCompliance(observationTimestamps: RecordedObservationTimestamp[], startMusterDate: Moment, endMusterDate: Moment): number {
```
richard3d, line 86: Between this method and the filtering above I get the sense we are doing a lot of stuff client-side (from DB perspective). Is it not possible to move some of this stuff directly into queries?

richard3d (changes requested):

packages/server/src/api/muster/muster.compliance.spec.ts
pfrollo: Good catch. I will adjust this code to what the UI requires today once I get there. I have no intention to change UI's behavior. If needed it can be done as part of a different ticket.

packages/server/src/api/muster/muster.compliance.ts
pfrollo: Correct, but I like the explicit style. ;-)

pfrollo: The order in this case doesn't matter but I like to keep the division last as it may result in some cases a grater precision of the result but it all depends on the numbers used.

packages/server/src/api/muster/muster.compliance.ts
pfrollo: Sure a different approach could have been taken but how do we know which one is better? What kind of criteria we would use?

packages/server/src/api/muster/muster.compliance.ts
```diff
@@ -45,29 +46,71 @@ export function toMusterIntermediateCompliance(rosters: Roster[]): IntermediateM
...
  */
 export type MusterCompliance = {
   totalMusters: number;
   mustersReported: number;
   musterPercent: number;
-} & IntermediateMusterCompliance;
+} & RosterForMusterCompliance;
 
 /**
  * Represents muster compliance without any compliance calculations
```
pfrollo, line 111: This is wrong. This is just a roster a subset of roster. This should be moved to the roster module.

packages/server/src/api/muster/muster.compliance.ts
pfrollo: See my comment above. Naming things is hard. I like `RosterEntry` the best but I don't want anyone to change its representation. I say, get your own roster data if you need it. Such ambiguous name invites others to do whatever they want with its representation. Oh well....

packages/server/src/api/muster/muster.postgres.controller.ts
pfrollo: Please see: https://app.asana.com/0/1188940305683068/1200529438257363
Namely: "For this effort we will only use the roster table, not the roster history table."
This was discussed with Austin. They typically only look at very recent data.

richard3d: I see, right on, we may need to revisit this with him because if they want exports of this data too (like the original muster page) the queries will have to be changed then in order to be accurate for any timespan.

packages/server/src/api/muster/muster.compliance.ts
richard3d: The general rule of thumb is that putting the responsibility on the DB will always be faster. For one the data is indexed, so you are not querying against values in a scanning/iterative fashion like here (at the very least things will be log(n) on the DB). The more items that are returned from the DB results in more memory being utilized on the server. Also in the case of a large number of rows being returned from the DB, your throughput will suffer as well (there is still a network connection from server to DB).

May not be a big deal since we are showing a limited number of days in the UI but if the data needs to be exported or larger queries need to be executed, it will not scale well. We can see what Austin says :)

pfrollo: I think it is hard to tell which method would be more efficient and easier to maintain especially looking at a small subset of the code.

packages/server/src/api/muster/muster.postgres.controller.ts
andersonad: Since there is no aggregation taking place over individuals to reach a unit level, but rather just filtering on a unit roster, I think for this initial PR it is fine to "ignore" the roster history table. 

It will be somewhat ambiguous to the user either way whether we include roster history, which means the time date range they chose is NOT the actual time date-range used for that user's calculations, or we exclude roster history, which will ensure the time date range matches, but that the user has a "low" value due to not being on the roster during that time-frame rather than through non-response. At present, I would say our users are "used to" the latter issue.

packages/server/src/api/muster/muster.compliance.ts
pfrollo: The performance improvements are now captured here: https://app.asana.com/0/1188940305683068/1200767809950586/f

packages/server/src/api/muster/muster.compliance.ts
pfrollo: done

packages/server/test.sh
```diff
@@ -21,6 +21,7 @@ while [[ "$#" -gt 0 ]]; do
   shift
 done
 
+export MUSTER_POSTGRES_CONTROLLER=true
```
richard3d, line 24: I thought both @rogelioii and @andersonad  said we don't need both? If so, let's get rid of this environment variable. Since you have already separated out the logic to a separate controller let's just replace the routes outright. That way if we need ES stuff again we can just wire up routes if we need to.

packages/server/src/api/muster/muster.router.ts
```diff
@@ -20,11 +19,12 @@ router.get(
   controller.getNearestMusterWindow,
 );
 
+const musterController = process.env.MUSTER_POSTGRES_CONTROLLER !== 'true' ? controller.getMusterRoster : musterPostgresController.getUserMusterCompliance;
```
richard3d, line 22: We really should remove this env variable this I feel like, see my other comments below.

packages/server/src/api/muster/muster.postgres.controller.it.ts
```diff
@@ -0,0 +1,81 @@
...
+    const toDate = '2020-01-14T23:00:00.000';
+
+    const res = await req.get(`/${orgId}/roster?fromDate=${fromDate}Z&toDate=${toDate}&unitId=${unitId}&page=0&limit=10`);
+
+    expectNoErrors(res);
+    expect(res.data)
+      .is
+      .eql(
+        { rows: [
+          { totalMusters: 10,
```
richard3d, line 36: Do we really need all this user information coming back? Like how does phone number and custom column for example even matter for the return of this calculation?

packages/server/src/api/muster/muster.postgres.controller.it.ts
```diff
@@ -0,0 +1,81 @@
...
+  let orgId: number;
+  let unitId: number;
+
+  beforeEach(async () => {
+    seedData = await seedAll();
+    req = new TestRequest(basePath);
+    req.setUser(await User.findOne());
+
+    orgId = seedData[0].org.id;
+    unitId = seedData[0].units.filter(((u: Unit) => u.name === 'Unit 1' && u.org!.id === orgId)).map((unit: { id: any }) => unit.id)[0];
```
richard3d, line 22: why `unit: {id: any}` in the map function? Why not just `unit: Unit` or at the very least make `id: number` ?

richard3d (changes requested):

packages/server/src/api/muster/muster.postgres.controller.it.ts
richard3d: Also using 'find' might be more appropriate than `map` so you return single value and don't need to index an array.

packages/server/test.sh
pfrollo: It needs to stay as we need to continue using the ES endpoint in Prod due to other work that needs to be done:
- custom fields 
- one off muster requirements
There are tickets for this work in Asana.

packages/server/src/api/muster/muster.router.ts
pfrollo: See my other comment.

packages/server/src/api/muster/muster.postgres.controller.it.ts
pfrollo: This is the original payload returned by the ES API. There was no change to the UI code at all.
Screen Shot 2021-08-12 at 9 42 37 AM
 The user information is displayed on the UI screen together with the calculations.

packages/server/src/api/muster/muster.postgres.controller.it.ts
pfrollo: simplified

pfrollo: good catch!

packages/server/src/api/muster/muster.postgres.controller.it.ts
richard3d: Got it, but I still do not see custom columns? IDK the payload may have just contained that initially due to how things have to be organized in ES.

richard3d (approved):

PR #194, opened 2021-08-10
Unit muster compliance table updates

richard3d:
Commit message says it all. Also contains fixes and cleanup suggested by @larskendall and @pfrollo.

### DIFF
--- a/packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
+++ b/packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
@@ -1,0 +1,222 @@
+import {
+  Grid,
+  Paper,
+  Table,
+  TableBody,
+  TableContainer,
+  TableCell,
+  TableHead,
+  TableRow,
+} from '@material-ui/core';
+
+import * as Plotly from 'plotly.js';
+import React, {
+  useCallback,
+  useEffect,
+  useState,
+} from 'react';
+import Plot from 'react-plotly.js';
+
+import { UnitSelector } from '../../../selectors/unit.selector';
+import { Modal } from '../../../actions/modal.actions';
+import { formatErrorMessage } from '../../../utility/errors';
+import { UserSelector } from '../../../selectors/user.selector';
+import { MusterClient } from '../../../client/muster.client';
+import { useAppDispatch } from '../../../hooks/use-app-dispatch';
+import { useAppSelector } from '../../../hooks/use-app-selector';
+
+export const UnitMusterComplianceTable = (props: any) => {
+  const dispatch = useAppDispatch();
+  const units = useAppSelector(UnitSelector.all);
+  const complianceBarChart = {
+    layout: {
+      height: 128,
+      autosize: true,
+      barmode: 'group',
+      yaxis: {
+        color: '#A9AEB1',
+        hoverformat: '.0f%',
+        showticklabels: true,
+        range: [0, 100],
+      },
+      xaxis: {
+        color: '#A9AEB1',
+        hoverformat: '',
+      },
+      hovermode: 'closest',
+      showlegend: false,
+      colorway: ['#7776AF', '#54BEA5', '#E87779', '#EACC73', '#81B2DA'],
+      margin: {
+        l: 40,
+        r: 20,
+        b: 20,
+        t: 20,
+      },
+    } as Partial<Plotly.Layout>,
+    config: {
+      displayModeBar: false,
+    } as Partial<Plotly.Config>,
+  };
+
+  const [rosterFromDateIso, setRosterFromDateIso] = useState(props.rosterFromDateIso);
+  const [rosterToDateIso, setRosterToDateIso] = useState(props.rosterToDateIso);
+  const [rosterUnitId, setRosterUnitId] = useState(props.rosterUnitId);
+  const [unitComplianceData, setUnitComplianceData] = useState<Map<string, ComplianceData>>(new Map());
+  const [unitCompliancePlotData, setUnitCompliancePlotData] = useState<Map<string, Plotly.Data[]>>(new Map());
+
+  const orgId = useAppSelector(UserSelector.org)!.id;
+
+  //
+  // Effects
+  //
+
+  const showErrorDialog = useCallback((title: string, error: Error) => {
+    void dispatch(Modal.alert(`Error: ${title}`, formatErrorMessage(error)));
+  }, [dispatch]);
+
+  const reloadMusterComplianceData = useCallback(async () => {
+    try {
+      console.log(units);
+      console.log(rosterUnitId);
+      console.log(rosterFromDateIso);
+      // determine if we are displaying for all units or just a single unit
+      const unitsToDisplay = rosterUnitId === -1 ? units : [units.find(unit => unit.id === rosterUnitId)];
+      const plotData = new Map();
+      const complianceDataByUnit = new Map();
+      for (const unit of unitsToDisplay) {
+        const xRange: string[] = [];
+        const yRange: number[] = [];
+        const colors: string[] = [];
+        let dayCount: number = 0;
+        let totalCompliance: number = 0;
+        const complianceData: ComplianceData = {
+          minCompliance: 100,
+          maxCompliance: 0,
+          avgCompliance: 0,
+        };
+        if (unit) {
+          const res = await MusterClient.getMusterComplianceByDateRange(orgId, unit.name, { isoStartDate: `${rosterFromDateIso}`, isoEndDate: `${rosterToDateIso}` });
+          if (res.musterComplianceRates.length > 0) {
+            res.musterComplianceRates.forEach(rate => {
+              const compliancePerc = rate.musterComplianceRate * 100;
+              xRange.push(rate.isoDate);
+              if (rate.musterComplianceRate !== null) {
+                yRange.push(compliancePerc);
+                colors.push('#7776AF');
+                // update min compliance
+                if (compliancePerc < complianceData.minCompliance) {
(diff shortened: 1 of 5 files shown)


packages/server/src/util/test-utils/seed.ts
```diff
@@ -183,25 +185,36 @@ async function generateOrg(admin: User, numUsers: number, numRosterEntries: numb
...
-    // hence the same timestamp for start and end
-    observations = observations.concat(
-      observationTestData(rosterEntries[i].edipi,
-        rosterEntries[i].unit.name,
-        reportSchemas[1],
-        moment('2020-01-02T10:00:00Z'),
-        moment('2020-01-02T10:00:00Z')),
-    );
-  }
+  const rosterEntriesByUnit = _.groupBy(rosterEntries, entry => entry.unit.name);
```
pfrollo, line 188: When you take a look at the `generateOrg()` function, you will see a pattern, calls to methods to generate data and then calling `save()`. However this replacement/addition does not follow the pattern. I think in either case this code would benefit from being extracted out into a function and depending on number of parameters, function name, clarity, it might be worthwhile adding a JSDoc, but that is to be seen once the code is extracted out.

packages/server/src/util/test-utils/seed.ts
```diff
@@ -183,25 +185,36 @@ async function generateOrg(admin: User, numUsers: number, numRosterEntries: numb
...
+  // for each unit...
+  _.forOwn(rosterEntriesByUnit, (value: Roster[]) => {
```
pfrollo, line 190: `value` represents `rosters` so maybe there is a better name better describing what it is?

packages/server/src/util/test-utils/seed.ts
```diff
@@ -183,25 +185,36 @@ async function generateOrg(admin: User, numUsers: number, numRosterEntries: numb
...
+    let count = 0;
+    // for each person on roster for the unit...
+    value.forEach((rosterEntry: Roster) => {
+      // we need to get the unit "number" and unfortunately there is no
+      // elegant way to do this currently aside from parsing it from the name
+      if (count < parseInt(rosterEntry.unit.name.split(' ')[1])) {
```
pfrollo, line 196: This `count` and the logic is confusing. How does count relate the the numerical part of the unit name? Some magic is happening here. 

I know what this does but only because I am familiar with unit names and some magic used to generated the seed data:
`parseInt(rosterEntry.unit.name.split(' ')[1])`
Maybe wrapping this into a method so that anyone new would not spend too much time trying to understand what is happening here? 

I might be missing something but would it be simpler for each roster (which has edipi & unit name) create two observations? It looks like that is what this code does. It would be a simple forEach loop over `rosterEntries`.

packages/server/src/api/muster/muster.controller.it.ts
```diff
@@ -17,31 +17,31 @@ describe(`Muster Controller`, () => {
...
     req.setUser(await User.findOne());
     // We don't' use a static id since ids may auto increment.
-    org1 = seedData[0].org.id;
+    org1 = seedData.find(data => data.org.indexPrefix === 'testgroup1').org.id;
   });
 
   describe(`${basePath}/:orgId/:unitName/complianceByDate : get`, () => {
     it(`should return full compliance with multiple active configs`, async () => {
-      const res = await req.get(`/${org1}/${unit1}/complianceByDate?isoDate=2020-01-02`);
+      const res = await req.get(`/${org1}/${unit5}/complianceByDate?isoDate=2020-01-02`);
```
pfrollo, line 25: Just like `org1` is an ID, I wonder whether `unit5` should also be an ID?

packages/server/src/api/muster/muster.controller.it.ts
```diff
@@ -50,12 +50,17 @@ describe(`Muster Controller`, () => {
...
+      const res = await req.get(`/${org1}/${unit5}/complianceByDateRange?isoStartDate=2020-01-03&isoEndDate=2020-01-04`);
       expectNoErrors(res);
 
       // The single config is active on jan 3 so we select the item in the list for testing
       const dayRate = res.data.musterComplianceRates.find((rate: GetMusterComplianceByDateResponse) => rate.isoDate.split('T')[0] === '2020-01-03');
       expect(dayRate.musterComplianceRate).equal(1);
     });
+
+    it(`should error with bad date input`, async () => {
+      const res = await req.get(`/${org1}/${unit5}/complianceByDateRange?isoStartDate=YYYY-MM-DD&isoEndDate=2020-01-04`);
```
pfrollo, line 62: Would the test fail for an invalid date? For example: 2020-99-03
I think that might be a better test then using `YYYY-MM-DD` as a value.
First I thought we are sending a date format, then I realized what is being tested here.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
```diff
@@ -0,0 +1,227 @@
...
+import { formatErrorMessage } from '../../../utility/errors';
+import { UserSelector } from '../../../selectors/user.selector';
+import { MusterClient } from '../../../client/muster.client';
+import { useAppDispatch } from '../../../hooks/use-app-dispatch';
+import { useAppSelector } from '../../../hooks/use-app-selector';
+
+export const UnitMusterComplianceTable = (props: any) => {
+  const dispatch = useAppDispatch();
+  const units = useAppSelector(UnitSelector.all);
+  const trendChart = {
```
pfrollo, line 31: `trendChart` is immediately consumed on line 62 but that is not evident unless one search for uses. I would get rid of it and construct the object on the line 62.  You could even move all this code to a separate module if you like.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const reloadMusterComplianceData = useCallback(async () => {
+    try {
+      console.log(units);
+      console.log(rosterUnitId);
+      console.log(rosterFromDateIso);
+      // determine if we are displaying for all units or just a single unit
+      const unitsToDisplay = rosterUnitId === -1 ? units : [units.find(unit => unit.id === rosterUnitId)];
+      const plotData = new Map();
+      const complianceDataByUnit = new Map();
+      for (const unit of unitsToDisplay) {
```
pfrollo, line 91: When I collapse this large block of code, then I see what it does. It "returns" these maps: `complianceDataByUnit` and `plotData`. I would prefer a function returning `complianceDataByUnit` and `plotData`. I would even move this function, a service, outside the the TSX module. 
Screen Shot 2021-08-11 at 11 24 27 AM

pfrollo (approved):
Approved with some comments.

packages/server/src/util/test-utils/seed.ts
```diff
@@ -183,25 +185,36 @@ async function generateOrg(admin: User, numUsers: number, numRosterEntries: numb
...
+        observations = observations.concat(
```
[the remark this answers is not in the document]
richard3d: Push will let you add multiple individual elements but concat marries two arrays together.

packages/server/src/util/test-utils/seed.ts
richard3d: The intent of this logic is to limit the number of users submitting observations and ultimately gain more compliance based on higher unit numbers. It does this based on the unit "number" which is equal to `parseInt(rosterEntry.unit.name.split(' ')[1])`. We use this number because IDs can't be trusted to be constant when running tests. <- I'll make sure to add this as a comment.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
richard3d: Good suggestion but I think I will leave as is for now. This logic is only used by this module and by this method. If that changes I will refactor then :)

packages/server/src/util/test-utils/seed.ts
richard3d: I think I misread the comment the first time...All that is happening here is that we are re-organizing the rosterEntries by their unit which is used to fill out values in a more sane manner for the observations. I do agree there is more refactoring to be done here though in general in seed.ts. I'll add this to that ticket.

packages/server/src/api/muster/muster.controller.it.ts
richard3d: Test will still fail but I can add that value case

packages/server/src/util/test-utils/seed.ts
richard3d: I think I misread the comment the first time...All that is happening here is that we are re-organizing the rosterEntries by their unit which is used to fill out values in a more sane manner for the observations. I do agree there is more refactoring to be done here though in general in seed.ts. I'll add this to that ticket.

packages/server/src/api/muster/muster.controller.it.ts
richard3d: Yep, I added a numeric value as part of the assertions for that test now too.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
richard3d: Good suggestion but I think I will leave as is for now. This logic is only used by this module and by this method. If that changes I will refactor then :)

pfrollo: As a learning conversation...Would it be a good idea for an API/service to do most of this heavy lifting instead of doing it in the UI client? The API would return appropriate and directly consumable data structures.

PR #196, opened 2021-08-11
Add saved filters.

larskendall:
https://app.asana.com/0/1188940305683068/1200592632584286

You should now be able to save filters on the roster table. Setting the filter to Custom will let you create a new filter from scratch, and can also be used as a private filter that will be persisted the same way filters worked before.

## Changes
- Refactored `QueryBuilder` to separate it into several files for the different components, since having everything in one file was getting cumbersome. The functionality was untouched for the most part, with just some renaming to make some of the components' names more explicit.
- Added `saved-filter` endpoints and client for accessing them.
- Added Save, Save As, and Delete options for roster filters.
- Added `SaveNewFilterDialog`.

  I used a more lightweight architecture for this dialog, which we might want to consider using for all of the dialogs in the future since it reduces the amount of props that need to be dynamically set and increases the flexibility of the dialog. The idea is to keep the dialog purely UI-focused, leaving all of the business logic up to the component rendering the dialog.

  We may even want to consider doing something similar for the modals in the future, since our current modal solution, though convenient, feels like it may be a bit of an anti-pattern.

## Screenshot
Screen Shot 2021-08-11 at 3 30 21 PM


AnthonyWhitaker (approved):
This looks awesome! Thanks for cleaning it up!

PR #198, opened 2021-08-12
Add support for exporting data to CSV from SQL for roster and org data.

richard3d:
Title says it all.

### DIFF
--- a/packages/server/src/api/export/export.elasticsearch.controller.ts
+++ b/packages/server/src/api/export/export.elasticsearch.controller.ts
@@ -26,7 +26,7 @@ import {
 import { Roster } from '../roster/roster.model';
 import { Unit } from '../unit/unit.model';
 
-class ExportController {
+class ExportElasticsearchController {
 
   async exportOrgToCsv(req: ApiRequest<null, null, ExportOrgQuery>, res: Response) {
     const startDate = moment(req.query.startDate ?? 0);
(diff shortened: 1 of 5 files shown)


packages/server/src/api/export/export.elasticsearch.controller.ts
```diff
@@ -185,4 +185,4 @@ class ExportController {
 
 }
 
-export default new ExportController();
+export default new ExportElasticsearchController();
```
pfrollo, line 188: Is this controller being used?

pfrollo: it is

packages/server/src/api/export/export.sql.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+  async exportOrgToCsv(req: ApiRequest<OrgParam, null, ExportOrgQuery>, res: Response) {
+    const startDate = moment(req.query.startDate ?? 0);
+    const endDate = moment(req.query.endDate);
+    const orgId = req.params.orgId;
+    const units: Unit[] = await Unit.createQueryBuilder('unit')
+      .select('unit.name')
+      .where(`unit.org_id = :orgId`, { orgId })
+      .getRawMany();
+
+    const unitNames: string[] = units.map(unit => unit.name);
```
pfrollo, line 25: Aren't unit names already returned on line 20 with the `select('unit.name')` query?

pfrollo (approved):
Sweet!

PR #199, opened 2021-08-13
Improvements and cleanup for unit muster compliance front-end

richard3d:
Wrap previous "waterfalled" API calls into single promise. Clean up output.

### DIFF
--- a/packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
+++ b/packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
@@ -28,6 +28,7 @@ import { useAppSelector } from '../../../hooks/use-app-selector';
 export const UnitMusterComplianceTable = (props: any) => {
   const dispatch = useAppDispatch();
   const units = useAppSelector(UnitSelector.all);
+  const orgId = useAppSelector(UserSelector.org)!.id;
   const complianceBarChart = {
     layout: {
       height: 128,
@@ -61,13 +62,11 @@ export const UnitMusterComplianceTable = (props: any) => {
   const [rosterFromDateIso, setRosterFromDateIso] = useState(props.rosterFromDateIso);
   const [rosterToDateIso, setRosterToDateIso] = useState(props.rosterToDateIso);
   const [rosterUnitId, setRosterUnitId] = useState(props.rosterUnitId);
-  const [unitComplianceData, setUnitComplianceData] = useState<Map<string, ComplianceData>>(new Map());
+  const [unitComplianceStats, setUnitComplianceStats] = useState<Map<string, ComplianceStats>>(new Map());
   const [unitCompliancePlotData, setUnitCompliancePlotData] = useState<Map<string, Plotly.Data[]>>(new Map());
 
-  const orgId = useAppSelector(UserSelector.org)!.id;
-
   //
-  // Effects
+  // Callbacks
   //
 
   const showErrorDialog = useCallback((title: string, error: Error) => {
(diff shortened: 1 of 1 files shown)


packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
```diff
@@ -61,13 +62,11 @@ export const UnitMusterComplianceTable = (props: any) => {
   const [rosterFromDateIso, setRosterFromDateIso] = useState(props.rosterFromDateIso);
   const [rosterToDateIso, setRosterToDateIso] = useState(props.rosterToDateIso);
   const [rosterUnitId, setRosterUnitId] = useState(props.rosterUnitId);
-  const [unitComplianceData, setUnitComplianceData] = useState<Map<string, ComplianceData>>(new Map());
+  const [unitComplianceStats, setUnitComplianceStats] = useState<Map<string, ComplianceStats>>(new Map());
```
larskendall, line 65: Is there any benefit to using Map here instead of a plain object since the keys are just strings? I guess seeing Map implies to me that something unusual is going on with non-string keys, but this seems to be doing the same thing as a regular object.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
```diff
@@ -76,69 +75,34 @@ export const UnitMusterComplianceTable = (props: any) => {
...
     try {
-      console.log(units);
-      console.log(rosterUnitId);
-      console.log(rosterFromDateIso);
       // determine if we are displaying for all units or just a single unit
       const unitsToDisplay = rosterUnitId === -1 ? units : [units.find(unit => unit.id === rosterUnitId)];
-      const plotData = new Map();
-      const complianceDataByUnit = new Map();
+      const plotDataByUnit = new Map();
+      const complianceStatsByUnit = new Map();
```
larskendall, line 81: These resolve to `Map<any, any>`, losing type checking. I guess this is a slight downside of Map, where it will sneak that through. You can add typings to the Map, but maybe it's better to just use plain objects here.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
```diff
@@ -155,6 +119,57 @@ export const UnitMusterComplianceTable = (props: any) => {
   // Functions
   //
 
+  const getUnitComplianceData = async (organizationId: number, unitName: string, isoStartDate: string, isoEndDate: string): Promise<ComplianceData> => {
+    const xRange: string[] = [];
+    const yRange: number[] = [];
+    const colors: string[] = [];
+    let dayCount: number = 0;
+    let totalCompliance: number = 0;
```
larskendall, line 127: I think it's fine in cases like these to just let TS infer the type. We've generally been using type inference where possible up to this point.

```suggestion
    let dayCount = 0;
    let totalCompliance = 0;
```

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
```diff
@@ -155,6 +119,57 @@ export const UnitMusterComplianceTable = (props: any) => {
...
+    const complianceStats: ComplianceStats = {
+      minCompliance: 100.0,
+      maxCompliance: 0,
+      avgCompliance: 0,
+    };
+    const res = await MusterClient.getMusterComplianceByDateRange(organizationId, unitName, { isoStartDate, isoEndDate });
```
larskendall, line 133: Probably `data` would be a clearer name for this variable, since it's not actually the response object.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
```diff
@@ -155,6 +119,57 @@ export const UnitMusterComplianceTable = (props: any) => {
...
+      res.musterComplianceRates.forEach(rate => {
+        const compliancePerc = rate.musterComplianceRate * 100;
+        xRange.push(rate.isoDate);
+        if (rate.musterComplianceRate !== null) {
+          yRange.push(compliancePerc);
+          colors.push('#7776AF');
+          // update min compliance
+          if (compliancePerc < complianceStats.minCompliance) {
+            complianceStats.minCompliance = compliancePerc;
+          }
```
larskendall, line 144: I wonder if the logic for this might be a little clearer/safer if we did something like this (outside of the loop)...

```typescript
      complianceStats.minCompliance = _.min(
        res.musterComplianceRates.map(x => x.musterComplianceRate * 100),
      );
```

After reading the code for a minute it was obvious what you were doing, but for a moment I was confused when I saw `minCompliance` defined higher than `maxCompliance` above. This way I'm suggesting is obviously less efficient since there would be multiple loops, but I doubt it would matter.

larskendall (changes requested):
Looks good! Just a couple minor suggestions.

The Map typing thing is the only thing that I think needs to be updated. The rest are just possible improvements.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
richard3d: it's `result` lol, but yeah vague af!

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
richard3d: We get the typing on the values also. Otherwise, how do I define an object where every value of a property/"key" must conform to a certain type?

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
richard3d: I'll add the typing, technically we get the check down in the code because we pass these as values to `setUnitComplianceStats` etc, which expect type conformity, I can make these temp variables explicit though.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
richard3d: Cool I think it makes sense to leave as is to prevent unnecessary iteration

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
larskendall: You should be able to do this to get the same level of typing with plain objects.

```typescript
const [unitComplianceStats, setUnitComplianceStats] = useState<{ [key: string]: ComplianceStats }>({});
```

larskendall: Or, alternatively, like this (though I generally prefer the first method since you can add context with a key name).

```typescript
const [unitComplianceStats, setUnitComplianceStats] = useState<Record<string, ComplianceStats>>({});
```

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
larskendall: I think since these default to `any` they'll actually pass those type checks as well. When a type is `any` it just passes all type checks. So you could potentially pass erroneous data through that wouldn't actually conform to the state values.

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
richard3d: Interesting I have never seen that notation for objects before. I gave it a try just for my own learning and the weird thing is that the value type is enforced but `[key: string]` seems to behave oddly. Like I can totally do the following (perhaps toString is called internally on number? idk seems sketchy though):
```
     const test: { [key: string]: string } = {};
      test[1] = 'whatev';
      test[2] = 'whatev';
 ```

File: packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx (hunk 2 above)
Comment (richard3d) [reply packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx:81]: Wow I thought .ts was a bit better at this, you are right! I don't even get a runtime error which is super disturbing even if the ambiguous map contains invalid data key or value data types. The lack of runtime error is actually a head scratcher for me...I'll have to brush up on my .ts a bit more it looks like. Either way the maps are typed now.

File: packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx (hunk 1 above)
Comment (larskendall) [reply packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx:65]: Yeah JS automatically converts numbers to strings in object keys. It would be nice if TS threw an error for that. Maybe there's some way to enforce that in the tsconfig.

It'll at least throw errors for union types, if you do something like this...

```typescript
type Things = 'a' | 'b';

const test: { [K in Things]?: number } = {};
test['a'] = 1; // ok
test['c'] = 2; // error
test[1] = 3; // error
```

packages/client/src/components/pages/muster-page/table-unit-muster-compliance.tsx
larskendall: We have the strict settings enabled, so implicit `any` types aren't allowed. But in this case they snuck through because they're just template types that default to `any`. The default map constructor looks like this: `Map<any, any>`.

That's one of the reasons I'm not sure how much I like Map. I also just sort of prefer staying in pure JSON as much of the time as possible, since that's the format that's being serialized and shuttled around everywhere.

PR #201, opened 2021-08-17
Fix "disableBackdropClick" deprecation warnings in frontend.

larskendall:
https://app.asana.com/0/1188940305683068/1200665371881023

This was starting to cause workflow issues with these warnings spamming the console, so I went ahead and fixed the problem.

Material UI no longer accepts a `disableBackdropClick` bool, and now forces you to handle it yourself in the `onClose` callback. Since there's no way to set a default for it in the theme anymore, and it would be a pain to have to handle this everywhere, I wrapped the Material UI Dialog component and disabled closing on backdrop click by default.

### DIFF
--- a/packages/client/src/components/dialog/dialog.tsx
+++ b/packages/client/src/components/dialog/dialog.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import {
+  Dialog as MuiDialog,
+  DialogProps,
+} from '@material-ui/core';
+
+/**
+ * Wrapper for Material UI Dialog class to allow us to set default behavior for
+ * onClose().
+ */
+export const Dialog = (props: DialogProps) => {
+  const { onClose, children, ...otherProps } = props;
+
+  const handleClose = (event: {}, reason: 'backdropClick' | 'escapeKeyDown') => {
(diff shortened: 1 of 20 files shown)


packages/client/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Wrapper for Material UI Dialog class to allow us to set default behavior for
+ * onClose().
+ */
+export const Dialog = (props: DialogProps) => {
+  const { onClose, children, ...otherProps } = props;
+
+  const handleClose = (event: {}, reason: 'backdropClick' | 'escapeKeyDown') => {
+    if (reason !== 'backdropClick') {
```
AnthonyWhitaker, line 15: Does this break functionality/accessibility for escapeKeyPressing?

richard3d (approved):

AnthonyWhitaker (approved):

packages/client/src/components/dialog/dialog.tsx
larskendall: Nope, using escape still works with this in place.

PR #204, opened 2021-08-19
Roster compliance refactor and one-time muster requirement calculation.

richard3d:
Title and commit message indicate the most important changes. There was an environment variable hanging out `MUSTER_POSTGRES_CONTROLLER` that was removed to ensure only postgres is used and ES could not still be accidentally used. Compliance has been refactored to match original design and organizational patterns used in the project. More tests will need to be written in the near future but this is currently minimum stable.


packages/server/src/api/muster/muster.controller.it.ts
```diff
@@ -20,25 +21,32 @@ describe(`Muster Controller`, () => {
     org1 = seedData.find(data => data.org.indexPrefix === 'testgroup1').org.id;
   });
 
-  describe(`${basePath}/:orgId/:unitName/complianceByDate : get`, () => {
+  describe(`${basePath}/:orgId/roster/complianceByDateRange : get`, () => {
     it(`should return full compliance with multiple active configs`, async () => {
-      const res = await req.get(`/${org1}/${unit5}/complianceByDate?isoDate=2020-01-02`);
+      const res = await req.get(`/${org1}/roster/complianceByDateRange?fromDate=2020-01-01&toDate=2020-01-03${pageParams}`);
```
larskendall, line 26: I think we can pass these query strings through using the req.get `config` argument like this. It should let you leverage type checking when building the query, and it'll make things a bit more readable.

```typescript
const query: MusterComplianceByDateRangeQuery = {
  fromDate: '2020-01-01',
  toDate: '2020-01-03',
}

const res = await req.get(`/${org1}/roster/complianceByDateRange`, {
  params: query,
});
```

One interesting thing to note, that I only recently discovered, is that type checking is more strict when using `: Type` vs `as Type`, which is why I define the query here before the request happens.

I hate how Axios takes a `params` property for the query string, but it'll get the job done anyways.

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -233,22 +237,110 @@ class MusterController {
...
+    const unitId = req.query.unitId!;
+    const orgId = req.appOrg!.id;
+
+    Log.info(`getUserMusterCompliance() for ${JSON.stringify({ orgId, unitId, fromDate, toDate })}`);
+
+    // get units for the org
+    const orgUnits = await Unit.createQueryBuilder('unit')
+      .select()
+      .where('unit.org_id = :orgId', { orgId })
+      .getMany();
```
larskendall, line 267: This style should be a little less error prone, and a little more readable I think. I generally only use the query builder as a fallback when something more advanced needs to happen.

```typescript
const orgUnits = await Unit.find({
  where: { org: orgId },
});
```

There is one bug with this style that Peter ran into, where the `Between` function doesn't seem to work correctly with timestamps. It looked like there was an open issue for it. Pretty lame, but other than that it should work well.

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -233,22 +237,110 @@ class MusterController {
...
+      unitNames.push(unit.name);
+      configsByUnitId.set(unit.id, unit.musterConfiguration);
+    });
+
+    // get the roster entries based on the unit IDs for the org
+    const rosterEntries = await Roster.createQueryBuilder('roster')
+      .leftJoinAndSelect('roster.unit', 'unit')
+      .select()
+      .where('roster.unit_id in (:...unitIds)', { unitIds })
+      .getMany();
```
larskendall, line 284:
```typescript
const rosterEntires = await Roster.find({
  relations: ['unit'],
  where: { unit: In(unitIds) },
});
```

Unrelated to this chunk of code, but if you ever need to get nested relations you can do something like this...

```typescript
const rosterEntires = await Roster.find({
  relations: ['unit', 'unit.org'],
  where: { unit: In(unitIds) },
});
```

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -233,22 +237,110 @@ class MusterController {
...
+
+      // get the compliant observation count for this current user/edipi
+      const userObservations = observationsByEdipi[rosterEntry.edipi];
+      complianceRecord.mustersReported = getCompliantUserObserverationCount(userObservations, configs);
+
+      // only perform percentage calculation if there are required musters,
+      // otherwise we default to 100% compliance.
+      if (complianceRecord.totalMusters > 0) {
+        complianceRecord.musterPercent = (complianceRecord.mustersReported / complianceRecord.totalMusters) * 100;
+      }
```
larskendall, line 327: You should be able to use the existing `calcMusterPercent` function here.

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -233,22 +237,110 @@ class MusterController {
...
+      .andWhere(`observation.edipi in (:...edipis)`, { edipis: rosterEdipis })
+      .andWhere(`observation.unit in (:...unitNames)`, { unitNames })
+      .andWhere(`observation.report_schema_org = :orgId`, { orgId })
+      .getRawMany();
+
+    const observationsByEdipi = _.groupBy(observations, obs => obs.edipi);
+    const complianceRecords: MusterCompliance[] = [];
+
+    // generate "compliance records" for each user
+    rosterEntries.forEach(rosterEntry => {
```
larskendall, line 305: I think the intent of the following chunk will be clearer if you use `map` here instead of `forEach`, and it also avoids mutations...

```typescript
const complianceRecords: MusterCompliance[] = _.map(rosterEntries, rosterEntry => {...});
```

packages/server/src/util/muster-utils.ts
```diff
@@ -577,6 +581,108 @@ function buildUnitStats(args: {
...
+}
+
+/**
+ * Gets the total required number of musters for a unit based on all configs for the unit
+ * over a given period of time.
+ * @param unitConfigs an array of config objects to check the count for the unit
+ * @param fromDate start of date range to get muster count for
+ * @param toDate cutoff for date range to get muster count for
+ */
+export function getUnitRequiredMusterCount(unitConfigs: object[] | undefined, fromDate: moment.Moment, toDate: moment.Moment): number {
```
larskendall, line 636: Do we have a more specific type than `object` we can use for `unitConfigs`? Can those be `MusterConfiguration[]`?

packages/server/src/util/muster-utils.ts
```diff
@@ -577,6 +581,108 @@ function buildUnitStats(args: {
...
+}
+
+/**
+ * Returns the number of compliant musters/observations for a given set of configs
+ * @param userObservations an array of {reportSchema, timestamp} objects for a single user
+ * @param unitConfigs an array of muster configurations for a single unit
+ */
+export function getCompliantUserObserverationCount(
+  userObservations: { report_schema_id: string; timestamp: string }[],
+  unitConfigs: object[] | undefined,
```
larskendall, line 656: Same as above comment regarding `object` type.

packages/server/src/util/muster-utils.ts
```diff
@@ -577,6 +581,108 @@ function buildUnitStats(args: {
...
+): number {
+  let complianceCount = 0;
+  if (userObservations) {
+    userObservations.forEach(userObs => {
+      // for each observation find out which config it should be related to
+      const config: any = unitConfigs?.find((c: any) => c.reportId === userObs.report_schema_id);
```
larskendall, line 662: We should have a type here instead of `any`. If we can give `unitConfigs` a more specific type then this will resolve itself (as long as `: any` is removed).

larskendall (changes requested):
Just a couple typing things that I'd say should definitely be fixed, and a few suggestions for simplifications.

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -233,22 +237,110 @@ class MusterController {
...
+      throw new BadRequestError('Invalid ISO date range.');
+    }
+
+    const rowLimit = parseInt(req.query.limit);
+    const pageNumber = parseInt(req.query.page);
+    // When unit id is missing, then data for all units are requested
+    const unitId = req.query.unitId!;
+    const orgId = req.appOrg!.id;
+
+    Log.info(`getUserMusterCompliance() for ${JSON.stringify({ orgId, unitId, fromDate, toDate })}`);
```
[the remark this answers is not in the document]
richard3d: IDK this was left over I think from Peter's original controller. Im going to remove.

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -233,22 +237,110 @@ class MusterController {
...
+    const toDate: moment.Moment = moment(req.query.toDate, 'YYYY-MM-DD');
+
+    if (!fromDate.isValid() || !toDate.isValid() || fromDate > toDate) {
+      throw new BadRequestError('Invalid ISO date range.');
+    }
+
+    const rowLimit = parseInt(req.query.limit);
+    const pageNumber = parseInt(req.query.page);
+    // When unit id is missing, then data for all units are requested
+    const unitId = req.query.unitId!;
```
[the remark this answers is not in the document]
richard3d: Good eye this was copy paste from Peter's original controller

packages/server/src/api/muster/muster.controller.ts
```diff
@@ -233,22 +237,110 @@ class MusterController {
...
+
+    // get unit information for the org
+    const unitIds: number[] = [];
+    const unitNames: string[] = [];
+    const configsByUnitId: Map<number, object[]> = new Map<number, object[]>();
+    orgUnits.forEach(unit => {
+      unitIds.push(unit.id);
+      unitNames.push(unit.name);
+      configsByUnitId.set(unit.id, unit.musterConfiguration);
+    });
```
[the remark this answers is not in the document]
richard3d: So the arrays are used as input to typeorm `In` function for queries below. I see this solves the usage of map but then we are performing unnecessary iteration, unless I am misunderstanding?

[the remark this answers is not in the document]
richard3d: Also as an aside, I've noticed a resistance to use `Map<>`, just curious on the rational for plain objects over something that provides a specific interface for key/value pairs in general?

PR #212, opened 2021-08-27
Expose expression validation capabilities to backend.

richard3d:
Expression values can now be evaluated on either client or server via single API. Also includes work for parameterized expressions.
https://app.asana.com/0/1188940305683068/1200856384181949/f

### DIFF
--- a/packages/client/src/components/query-builder/query-builder-row.tsx
+++ b/packages/client/src/components/query-builder/query-builder-row.tsx
@@ -14,6 +14,7 @@ import {
   getFieldDefaultQueryValue,
   getFieldQueryOpDesc,
   isArrayOp,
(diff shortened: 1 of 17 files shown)


richard3d:
I think originally I was unaware if the "saved-filter" stuff would be in at this point or if you just needed API to call it and then were going to incorporate in your code. I'll update PR.

richard3d:
I made one more push to fix an issue with the column name calculations for expression references

larskendall (approved):
I'm not sure if we need the expression ref stuff (I thought we decided not to go that route for now), but since time is of the essence right now, and it doesn't seem to be hurting anything, I say go ahead and merge and we can come back to it if necessary.

PR #213, opened 2021-08-30
Use client Count API rather than obtain all records and counting client-side.

richard3d:
Small PR per observations from @larskendall, title says it all.

### DIFF
--- a/packages/client/src/components/pages/home-page/home-page.tsx
+++ b/packages/client/src/components/pages/home-page/home-page.tsx
@@ -19,6 +19,7 @@ import ArrowUpwardIcon from '@material-ui/icons/ArrowUpward';
 import FavoriteIcon from '@material-ui/icons/Favorite';
 import clsx from 'clsx';
 import moment from 'moment';
+import { OrphanedRecordSelector } from '../../../selectors/orphaned-record.selector';
 import { UserSelector } from '../../../selectors/user.selector';
 import { WorkspaceSelector } from '../../../selectors/workspace.selector';
 import { Link } from '../../link/link';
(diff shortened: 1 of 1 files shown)


AnthonyWhitaker (approved):

packages/client/src/components/pages/home-page/home-page.tsx
```diff
@@ -138,20 +135,10 @@ export const HomePage = () => {
...
-            limit: '',
-            unit: undefined,
-          },
-        }));
-      } catch (error) {
-        void dispatch(Modal.alert('Get Orphaned Records', formatErrorMessage(error, 'Failed to get orphaned records')));
-      }
+      }    
+      const orcResponse = await OrphanedRecordClient.getOrphanedRecordsCount(orgId);
+      setOrphanedRecordCount(orcResponse.count);
```
larskendall, line 140: This data is actually already in the redux store, and can be refreshed via an action. So you shouldn't need to store it as state in the home page. Here's an example of what that would look like...

```typescript
void dispatch(OrphanedRecordActions.fetchCount({ orgId: orgId! }));
```

packages/client/src/components/pages/home-page/home-page.tsx
```diff
@@ -121,12 +117,13 @@ export const HomePage = () => {
...
-  const orphanedRecords = useAppSelector(OrphanedRecordSelector.root);
-  const [accessRequests, setAccessRequests] = useState<ApiAccessRequest[]>([]);
-  const [musterComplianceLastTwoWeeks, setMusterComplianceLastTwoWeek] = useState<number[]>([1.0, 0.0]);
   const favoriteDashboards = useAppSelector(UserSelector.favoriteDashboards)!;
   const workspaces = useAppSelector(UserSelector.workspaces)!;
   const dashboards = useAppSelector(WorkspaceSelector.dashboards)!;
+  
+  const [accessRequests, setAccessRequests] = useState<ApiAccessRequest[]>([]);
+  const [musterComplianceLastTwoWeeks, setMusterComplianceLastTwoWeek] = useState<number[]>([1.0, 0.0]);
+  const [orphanedRecordCount, setOrphanedRecordCount] = useState<number>(0);
```
larskendall, line 126: These types will be automatically inferred automatically. Just a heads up.

larskendall (changes requested):

richard3d:
@larskendall can you confirm this is good now?

packages/client/src/components/pages/home-page/home-page.tsx
```diff
@@ -140,6 +141,7 @@ export const HomePage = () => {
       }
 
       try {
+        await dispatch(OrphanedRecordActions.fetchCount({ orgId: orgId! }));
         await dispatch(OrphanedRecordActions.fetchPage({
```
larskendall, line 145: Can this `fetchPage` call be removed? It looks like that may have been there as a means of getting `totalRowsCount` for the orphaned records count, which shouldn't be necessary now.

larskendall (changes requested):

packages/client/src/components/pages/home-page/home-page.tsx
richard3d: done

PR #216, opened 2021-08-31
Completed controller review and model decorations for indexing.  

mbrinkg8tr:
Completed controller review and model decorations for indexing.  Generated migration scripts, completed local testing.  Satisfies Asana task: 
[Add indices for columns we're querying against (Postgres)](https://app.asana.com/0/1188940305683068/1200218215182686/f)

### DIFF
--- a/packages/server/src/api/access-request/access-request.model.ts
+++ b/packages/server/src/api/access-request/access-request.model.ts
@@ -8,6 +8,7 @@ import {
   Column,
   BeforeInsert,
   BeforeUpdate,
+  Index,
 } from 'typeorm';
 import { formatPhoneNumber } from '@covid19-reports/shared';
 import { Org } from '../org/org.model';
@@ -53,6 +54,7 @@ export class AccessRequest extends BaseEntity {
   })
   whatYouDo!: string[];
 
+  @Index('access-request-sponsorName')
   @Column()
   sponsorName!: string;
 
--- a/packages/server/src/migration/1630344306894-DatabaseIndexOptimization.ts
+++ b/packages/server/src/migration/1630344306894-DatabaseIndexOptimization.ts
@@ -1,0 +1,39 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class DatabaseIndexOptimization1630344306894 implements MigrationInterface {
+
+  name = 'DatabaseIndexOptimization1630344306894';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "user_notification_setting" ALTER COLUMN "last_notified_date" SET DEFAULT null`);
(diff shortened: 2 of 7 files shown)


packages/server/src/api/access-request/access-request.model.ts
```diff
@@ -53,6 +54,7 @@ export class AccessRequest extends BaseEntity {
   })
   whatYouDo!: string[];
 
+  @Index('access-request-sponsorName')
```
richard3d, line 57: I'm no so familiar with typeORM indices but @larskendall can you confirm if this annotation is even necessary if we are defining in the migration already (see below)?

mbrinkg8tr: The migration was computed based on the definition of the models.  Removing the index ornamentation would break ORM/Schema consistency, but would not affect functionality of the code.  In performing some local testing, all tests continue to pass (with ornamentation removed, and migration .ts in place), and there didn't seem to be any significant performance changes.  Plus, removal of the indices post migration generation might result in removal of the indices if future ORM/schema changes are implemented.

larskendall: The decorator is necessary for TypeORM to generate schema changes. If we don't have them I think TypeORM will constantly try to remove those indexes any time you generate a new migration.

packages/server/src/migration/1630344306894-DatabaseIndexOptimization.ts
```diff
@@ -0,0 +1,39 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class DatabaseIndexOptimization1630344306894 implements MigrationInterface {
+
+  name = 'DatabaseIndexOptimization1630344306894';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "user_notification_setting" ALTER COLUMN "last_notified_date" SET DEFAULT null`);
```
larskendall, line 8: There's a bug in TypeORM unfortunately where it always wants to alter this column, even if it already did in the last migration. So it can safely be removed. There's a matching line in the `down` function as well that can be removed.

larskendall (commented):
I might opt for leaving out the custom index names in the decorators, since that would add some extra work if we ever refactor these field names and could leave things in an awkward state if we forget.

Is there any particular benefit you've found with using those custom index names?

larskendall:
> This is looking pretty good!
> 
> I might opt for leaving out the custom index names in the decorators, since that would add some extra work if we ever refactor these field names and could leave things in an awkward state if we forget.
> 
> Is there any particular benefit you've found with using those custom index names?

Discussed this with Martin and we agreed the custom index names are a good way to go for the sake of readability in the actual database schema.

larskendall (approved):

PR #217, opened 2021-09-01
Use redux-toolkit api for entities in views.

larskendall:
https://app.asana.com/0/1188940305683068/1200849990119126

I refactored the View and ViewLayout components fairly substantially, since the RTK api stuff made it possible to simplify a lot of things and sort of required a lot of structure to change. One of the benefits is you only need to pass the entity type to the view layout now instead of specific endpoints. It will automatically find the associated endpoints for that entity type. We may be able to simplify it even more by having a single set of entity endpoints, but this should work well for now and be easily swappable for that architecture if we choose to go that route.

I pulled a lot of the code that was in ViewLayoutSelector up to ViewLayout, since it seemed like it made some things a little cleaner and easier to work with. That selector component could potentially be made even more generic to be reusable for other purposes, since it's not really dependent on any layout stuff at this point.

There's one bug I noticed where the layout selector delete button only works if you currently have that layout selected. I'll fix it separately after this PR.

Things are in a little bit of a weird state right now with the RTK api system being used in some places and not in others. I'd really like to make a pass across the entire app to switch over to using RTK, but I wanted to get this stuff in to unblock you. It's looking like RTK is going to really help clean things up. Its caching and auto-refetching system handles a _lot_ of stuff that we've been doing manually over and over up to this point. So I think once we switch over to it fully we'll be able to strip out a whole lot of code.


AnthonyWhitaker (approved):
Works on my machine! (more of a functional test on top of a light code review)

PR #224, opened 2021-09-08
Fix observations, again

AnthonyWhitaker:
Should be a quick one. The main goal is to get all the report data into `observation.customColumns`

### DIFF
--- a/packages/server/src/api/observation/observation.controller.ts
+++ b/packages/server/src/api/observation/observation.controller.ts
@@ -1,7 +1,7 @@
 import { Request, Response } from 'express';
 import { EntityManager, getConnection } from 'typeorm';
-import { RosterInfo } from '@covid19-reports/shared';
-import { ObservationApiModel, ObservationApiRawReportModel } from './observation.types';
+import { CustomColumns, RosterInfo } from '@covid19-reports/shared';
+import { ObservationApiModel, Report } from './observation.types';
 import { Observation } from './observation.model';
 import { ApiRequest, EdipiParam } from '../api.router';
 import { Log } from '../../util/log';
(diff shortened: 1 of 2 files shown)


packages/server/src/api/observation/observation.controller.ts
```diff
@@ -80,39 +79,30 @@ function hasReportingGroup(rosterInfo: RosterInfo, reportingGroup: string | unde
...
   const observation = new Observation();
+  const { ID, EDIPI, ReportingGroup, ReportType, Timestamp, ...customColumns } = req.body;
   observation.documentId = req.body.ID;
   observation.edipi = req.body.EDIPI;
   observation.timestamp = timestampColumnTransformer.to(req.body.Timestamp);
   observation.reportSchema = reportSchema;
-  observation.unit = req.body.Details?.Unit ?? '';
+  observation.unit = `${req.body.Details?.Unit ?? ''}`;
   observation.reportingGroup = getReportingGroup(req);
+  observation.customColumns = customColumns as unknown as CustomColumns;
```
AnthonyWhitaker, line 105: @larskendall any thoughts on this? technically CustomColumns isn't the right type due to nested objects, and arrays

larskendall: Do you suppose we should flatten that data down before setting it on `observation.customColumns`? If you think we're ok to have nested data in there maybe we could just define CustomColumns as a recursive type sort of like this:

```typescript
export type CustomColumns = {
  [columnName: string]: ColumnValue | CustomColumns;
};
```

However, I'd probably lean toward flattening it if it doesn't cause issues, just because it seems like it'll be a headache to worry about nested types everywhere we interact with custom columns. We could potentially generate new keys for a flat object similarly to what the CSV export does where it just joins the nested keys with a `.`. We wouldn't lose any of the structure that way and could unflatten it any time we wanted.

larskendall (commented):
Looks good! Just one question and thoughts on your comment.

packages/server/src/api/observation/observation.controller.ts
rogelioii: How would a flattened custom columns affect future creation of saved filters related to custom column content?  Say, I want to have a saved filter for all individuals that show a temperature greater than 100 ?  In my mind, this may be problematic when flattened?

PR #227, opened 2021-09-10
Query builder null handling

AnthonyWhitaker:
image


packages/client/src/utility/query-builder-utils.ts
```diff
@@ -104,12 +112,16 @@ export function getFieldDefaultQueryOp(field: QueryField): QueryOp {
   return Object.keys(opDesc)[0] as QueryOp;
 }
 
+export function opRequiresValue(op: QueryOp) {
```
rogelioii, line 115: don't you need to consider between in this case?

rogelioii (commented):
Looks good to me.

PR #228, opened 2021-09-10
Update muster configurations to be based on saved filters.

jdgarrett:
This PR adds new database models for muster configurations and muster filters.  Generally the muster_filters table would just be a many to many relationship, but it has been separated to have it's own model in case we need to supply parameters to the filter in the future.

This PR also adds a new muster configuration page to the UI which allows the creation of muster configurations and allows the user to assign filters to that configuration.  The Units page has been greatly simplified with all of the muster configuration logic removed.


packages/server/src/api/muster/muster.controller.ts
```diff
@@ -449,7 +620,7 @@ export async function getRosterMusterComplianceRecordsByDateRange(
   requestedUnits.forEach(unit => {
     unitIds.push(unit.id);
     unitNames.push(unit.name);
-    configsByUnitId.set(unit.id, unit.musterConfiguration);
+    // configsByUnitId.set(unit.id, unit.musterConfiguration);
```
larskendall, line 623:
```suggestion
```

larskendall (commented):
Looks good as far as I can tell. It all seemed to work when I tested it. I only had a few comments for some potential improvements, but feel free to address any you agree with and then merge when you'd like.

The one that's probably most worth addressing is the confirmation modal simplification, since it will reduce the complexity of the component a good bit.

PR #231, opened 2021-09-14
Fix errors when using 'between' in filters.

larskendall:
https://app.asana.com/0/1188940305683068/1200983953900253

This was an issue where the row value needed to be converted into an array for the range value editor. I went ahead and cleaned up the range value editor a bit as well since it was pretty difficult to understand.

Query rows currently only support a single expression per row, so I disabled expressions for range filters for now.


richard3d:
Looks good to me 👍

richard3d (approved):

PR #235, opened 2021-09-16
feat(everywhere): remove workspaces, kibana, dashboards

nathleia:


packages/server/src/api/muster/muster.controller.ts
```diff
@@ -130,56 +127,56 @@ class MusterController {
...
-    const unitTrends = await getMusterUnitTrends({
-      userRole: req.appUserRole!,
-      currentDate,
-      weeksCount,
-      monthsCount,
-    });
-
-    res.json(unitTrends);
-  }
+  // async getMusterRoster(req: ApiRequest<OrgRoleParams, null, GetMusterRosterQuery>, res: Response<Paginated<Partial<Roster>>>) {
```
richard3d, line 130: Hmmm....Not sure why this is commented out? What is the implication with workspaces here?

jdgarrett: I think a lot of stuff in this PR won't make sense until the muster compliance task is finished.  I have re-written most of the muster compliance logic.

richard3d: alright I'll defer to @jdgarrett then since he is working that too

richard3d: alright I'll defer to @jdgarrett then since he is working that too

PR #241, opened 2021-09-22
Fix to prevent camel case field from entering a query string

mbrinkg8tr:
This approach seemed a bit hacky to me, I'm open to suggestions.


mbrinkg8tr:
https://app.asana.com/0/1200796464156567/1201020649932884

larskendall (approved):
I think this looks good. I'm not sure if there's a very simple less-hacky solution, since our database models all use camel case and TypeORM converts everything to snake case internally. This should work fine for now.

PR #242, opened 2021-09-22
fix(observations): add missing deps to useEffect

nathleia:
skipped the ones in use-effect-debounced.ts & query-builder.ts because these will break things.


mbrinkg8tr (approved):
Some wierdness with roster-page vs. roster-page-2, but I suspect this will be resolved in future update.

PR #247, opened 2021-09-28
Update muster compliance to use observation table

jdgarrett:
This pull requests adds muster information to observations that will get populated when an observation is created.  The observation will be associated with a muster window and a roster history entry.

Non-reporting individuals will be added to the observations table when the appropriate endpoint is called in the API and the closed muster windows are computed.

Muster compliance calculations have been updated to use the observation table and old muster logic and endpoints have been removed.  Muster controller and muster util tests have also been removed as the tests were no longer valid with the updated logic.

Once this PR is merged, the associated PR on the processor will also need to be merged: https://github.com/mystatus/se-ingest-processor/pull/27


richard3d (approved):

larskendall (approved):
Did a very rough skimming since I'm short on time, but it looks good overall to me.

PR #249, opened 2021-09-29
Enables patching of records using the action system.

mbrinkg8tr:


packages/client/src/entity-actions/edit-column-action.tsx
```diff
@@ -122,6 +125,7 @@ export function registerActionsForUpdatableColumns(columns: ColumnInfo[], entity
       canMenu: action.renderType !== 'editor',
       refetchEntities: true,
     },
+    // eslint-disable-next-line require-await
```
larskendall, line 128: Looks like this can be removed.

packages/client/src/entity-actions/edit-column-action.tsx
```diff
@@ -130,9 +134,13 @@ export function registerActionsForUpdatableColumns(columns: ColumnInfo[], entity
...
-
-      // entityApi.patch(row.id, patchPayload)
-      console.log({ patchPayload, row });
+      const orgId = UserSelector.orgId(store.getState())!;
+      const endpoint = entityApi[entityType].endpoints.patch;
+      await store.dispatch(endpoint.initiate({
+        orgId,
+        rowId: row.id,
+        payload: patchPayload,
+      }));
```
larskendall, line 143: I think it may be better to pass in the `patch` function after setting up the hook in the caller's component. That way the caller can easily listen for errors if necessary, and the behavior might be a little more transparent. It's probably also not a best practice to access the store directly like this outside of an action.

So it might look something like this instead of passing in the entity type...

```typescript
  const [patch, {
    error: patchError,
  }] = entityApi[entityType].usePatchMutation();

  useEffect(() => {
    registerActionsForUpdatableColumns(allowedColumns, patch);
  }, [allowedColumns, patch]);
```

packages/server/src/api/roster/roster.controller.ts
```diff
@@ -320,6 +320,17 @@ class RosterController {
     res.json(entry);
   }
 
+  async patchRosterEntry(req: ApiRequest<OrgRosterParams, RosterEntryData>, res: Response) {
+    assertRequestParams(req, ['rowId']);
+    const entryId = +req.params.rowId;
```
larskendall, line 325: We may want to do a `parseInt` here and then check to make sure it's not NaN. Otherwise it looks like we'll get a SQL error.

packages/client/src/entity-actions/roster/addRosterEntry.tsx
```diff
@@ -45,18 +45,19 @@ export const addRosterEntry = new BulkAction(
     pii: false,
     refetchEntities: true,
   },
+  // eslint-disable-next-line require-await
   (async () => {
     return defer();
   }),
```
larskendall, line 51: We could avoid disabling eslint here by doing this instead. There are a couple other chunks like this I noticed that would be worth switching as well if you can.

```typescript
  () => {
    return Promise.resolve(defer());
  },
```

larskendall (changes requested):
Looks pretty good! I think the main comment to address is regarding having a standard entity endpoint for patch. Otherwise it's mainly just minor suggestions.

packages/client/src/api/api-utils.ts
```diff
@@ -80,6 +83,20 @@ function buildEntityEndpoints<TEntityData>(
...
+      rowId: number;
+      payload: Record<string, ColumnValue>;
+    }>({
+      query: ({ orgId, rowId, payload }) => ({
+        url: `${orgId}/${rowId}`,
+        method: 'PATCH',
+        body: payload,
+      }),
+      invalidatesTags: [entityType],
     }),
```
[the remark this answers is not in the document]
mbrinkg8tr: yea I agree, I'll rename it - and I didn't implement any other endpoints in this fix as I wasn't sure about the path to genericism at that time.  I'll do that as well.

packages/client/src/entity-actions/edit-column-action.tsx
mbrinkg8tr: This actually breaks a few of the hooks rules, which is why we took the above approach (see code in context).

packages/server/src/api/roster/roster.controller.ts
mbrinkg8tr: Yup!

packages/client/src/entity-actions/roster/addRosterEntry.tsx
mbrinkg8tr: Cool! Will implement

packages/client/src/entity-actions/edit-column-action.tsx
mbrinkg8tr: Updating this to use hooks requires refactoring that may be outside the scope of this task.

packages/client/src/api/api-utils.ts
```diff
@@ -80,6 +83,20 @@ function buildEntityEndpoints<TEntityData>(
...
+      orgId: number;
+      rowId: number;
+      payload: Record<string, ColumnValue>;
+    }>({
+      query: ({ orgId, rowId, payload }) => ({
+        url: `${orgId}/${rowId}`,
+        method: 'PATCH',
+        body: payload,
+      }),
+      invalidatesTags: [entityType],
```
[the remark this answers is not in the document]
mbrinkg8tr: type Paginated<> does not contain the property 'map', preventing the enumeration of Tags per entity.

packages/client/src/api/api-utils.ts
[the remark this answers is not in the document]
mbrinkg8tr: Attempting to add "patchEntities" to entity-utils.ts in the EntityController broke migrations.  Might need some help getting this in.

mbrinkg8tr:
Rejected

PR #250, opened 2021-09-30
Add endpoints for notification alert data.

richard3d:
This PR takes the calculations previously performed in the notification engine and adds them as endpoints that return the data for the given alerts. The body of the return is slightly different since it maps the users who need to be modified with the data for that given user/setting.
e.g.
```
[
    {
       userSetting: ...,
       alertData:....
    },
    ...
]
```
This code updates the daily count and last notified columns of the user settings however it does not reset daily counts. It is assumed the notification engine is still responsible for sending the alerts to users as well.


packages/server/src/api/notification/notification-alert.controller.ts
```diff
@@ -0,0 +1,314 @@
...
+      .filter((setting: UserNotificationSetting) => {
+        return isEligibleForNotification(setting);
+      });
+
+    // for each user we need to see if it is time to fire an alert and calculate the values
+    for (const setting of userSettings) {
+      const alertData: AlertData = {};
+      const minutes = getMinutesSinceLastNotification(setting);
+      const rows = await Observation.createQueryBuilder('observation')
+        .select('COUNT(observation.unit) as count, observation.unit')
```
jdgarrett, line 37: One thing that is a bit weird about the observation table is that `observation.unit` is actually a user-reported field.  The real unit should actually be acquired by joining the observation table with roster history on `roster_history_entry_id` and fetching the unit ID from there.  I'm not sure which one this notification is supposed to be based on, but I would expect it to be the one from roster history so that it matches the target user's permissions.  I am guessing the final result also needs to be filtered by what units the user has access to.

packages/server/src/api/notification/notification-alert.controller.ts
```diff
@@ -0,0 +1,314 @@
...
+        .where('observation.report_schema_org = :org', { org: req.appOrg!.id })
+        .andWhere(`observation.report_schema_id = 'es6ddssymptomobs'`)
+        .andWhere(`observation.custom_columns->>'TalkToSomeone' = 'true'`)
+        .andWhere('observation.timestamp BETWEEN :minTime AND :maxTime', {
+          minTime: moment().subtract(minutes, 'minutes'),
+          maxTime: moment(),
+        })
+        .groupBy('observation.unit')
+        .getRawMany();
+      console.log(rows);
```
jdgarrett, line 47: Remove `console.log`.

packages/server/src/api/notification/notification-alert.controller.ts
```diff
@@ -0,0 +1,314 @@
...
+          `((observation.custom_columns->>'Score')::int >= 20 
+          OR (observation.custom_columns->>'TemperatureFahrenheit')::int >= 102
+          OR observation.custom_columns->>'Category' In (:...keywords))`,
+          {
+            keywords: ['medium', 'high', 'very High'],
+          },
+        )
+        .groupBy('observation.unit')
+        .getRawMany();
+      rows.forEach((row: any) => {
```
jdgarrett, line 98: Same as above, this should probably be using the roster history unit, and be filtered based on user permissions.

packages/server/src/api/notification/notification-alert.controller.ts
```diff
@@ -0,0 +1,314 @@
...
+    for (let lookBackIndex = lookBackDays; lookBackIndex > 0; lookBackIndex--) {
+      const rows = await Observation.createQueryBuilder('observation')
+        .select(`observation.custom_columns->>'Symptoms' as symptoms, observation.unit`)
+        .where('observation.report_schema_org = :org', { org: req.appOrg!.id })
+        .andWhere(`observation.report_schema_id = 'es6ddssymptomobs'`)
+        .andWhere('observation.timestamp BETWEEN :minTime AND :maxTime', {
+          minTime: moment().subtract(lookBackIndex, 'days'),
+          maxTime: moment().subtract(lookBackIndex - 1, 'days'),
+        })
+        .orderBy('observation.unit')
```
jdgarrett, line 152: Same as above, select from roster history and filter by user permissions for the final alert.

packages/server/src/api/notification/notification-alert.controller.ts
```diff
@@ -0,0 +1,314 @@
...
+    res.json(numTrendingSymptomsByUnit);
+  }
+
+  // return the number of symptoms per unit that are trending upwards or downwards over last 3 days
+  async getLowMusterRates(req: ApiRequest<OrgSettingParams>, res: Response) {
+    const output: OutputRecord[] = [];
+
+    const stats = await musterComplianceStatsByDateRange(
+      req.appOrg!,
+      req.appUserRole!,
```
jdgarrett, line 226: Since this endpoint is called by the notification service, I believe this will be full admin permissions, which should be fine, but it's something to keep in mind.  Only units that match the user's `UserRole` should also be part of the alert for each user as well.

packages/server/src/api/notification/notification.router.ts
```diff
@@ -50,4 +51,39 @@ router.put(
   controller.updateUserNotificationSetting,
 );
 
+router.get(
+  '/:orgId/users-requesting-medical-attention',
+  bodyParser.json(),
+  requireOrgAccess,
```
jdgarrett, line 57: These should probably all have `requireInternalUser` in addition to `requireOrgAccess` to make sure only internal applications are able to use this endpoint.

packages/server/src/util/test-utils/data/observation-generator.ts
```diff
@@ -13,6 +14,18 @@ export function observationTestData(userEdipi: string, unitName: string, reportS
...
     observation.documentId = `DocumentId_${userEdipi}_${reportSchema.id}_${count}`;
     observation.edipi = userEdipi;
+    const customCol: CustomColumns = {
+      Details: {
+        TalkToSomeone: true,
+        TemperatureFahrenheit: 102,
+        Symptoms: 'cough, fever',
+      },
+      Category: {
+        Keyword: 'Medium',
```
jdgarrett, line 24: I think category shouldn't be coming in with `Keyword`.  I believe that's something specific to elasticsearch.  Instead it should probably be `Category: 'Medium',`.

packages/server/src/api/notification/notification-alert.controller.ts
```diff
@@ -0,0 +1,314 @@
...
+        .andWhere('observation.timestamp BETWEEN :minTime AND :maxTime', {
+          minTime: moment().subtract(minutes, 'minutes'),
+          maxTime: moment(),
+        })
+        .andWhere(
+          `((observation.custom_columns->>'Score')::int >= 20 
+          OR (observation.custom_columns->>'TemperatureFahrenheit')::int >= 102
+          OR observation.custom_columns->>'Category' In (:...keywords))`,
+          {
+            keywords: ['medium', 'high', 'very High'],
```
jdgarrett, line 93: I believe these should all be lower case and the result from `Category` should be wrapped in `LOWER` to ensure that differences in casing don't let things fall through the cracks.  There seems to be inconsistency in the sample documents, so it's best to play it safe.

packages/shared/src/entity.types.ts
```diff
@@ -1,6 +1,6 @@
 import moment from 'moment-timezone';
 
-export type ColumnValue = string | boolean | number | null;
+export type ColumnValue = object | string | boolean | number | null;
```
jdgarrett, line 3: In what case is `ColumnValue` an `object`?  I am guessing this is here due to `Category` having a `Keyword` still, but it would be great if we could avoid it, since it could be a bit too open.

jdgarrett (changes requested):
I haven't tested these endpoints locally yet, but I noticed some issues with the handling of observation units and user permissions that should be addressed prior to merge.

packages/server/src/api/notification/notification-alert.controller.ts
richard3d: I am a little lost on the permissions comment, my understanding is these endpoints are only ever called from the notification engine.

packages/server/src/api/notification/notification.router.ts
richard3d: Ah ok this makes sense, thanks!

packages/server/src/util/test-utils/data/observation-generator.ts
richard3d: good catch thanks!

packages/shared/src/entity.types.ts
richard3d: I had to use this as a hack while you were dealing with the nested JSON in custom columns. Glad you caught this though, I removed it.

packages/server/src/api/notification/notification-alert.controller.ts
richard3d: Also regarding joining with roster history - doesn't it actually go a bit further? The alert data is a dictionary of units (names in this case) and the counts. So I would have to join with unit also to actually get the name. Should this be two separate queries where I join with roster entry id and then back fill the names by looking up the resulting IDs?

packages/server/src/util/test-utils/data/observation-generator.ts
```diff
@@ -0,0 +1,22 @@
+import { Observation } from '../../../api/observation/observation.model';
```
larskendall, line 1: It looks like this file got added back in, but the function isn't used. I had pulled it to avoid confusion (since I fell victim to it for a moment). It should be safe to delete.

larskendall (approved):
I just skimmed a lot of the code since there were some pretty complex algorithms, and I'm guessing most of that has been ported anyways. There was only one very minor note, which you can feel free to ignore if you'd prefer.

jdgarrett (approved):

PR #252, opened 2021-10-01
Fix and standardize fully qualified column names.

larskendall:
https://app.asana.com/0/1188940305683068/1200982891588110

This should bring some order to the way column selects were being generated ad-hoc in lots of places, and resolve any other bugs with filter queries not working or throwing errors. I tried testing all of the column types I could in filters for both the observations and roster pages, and they all appeared to work.

I moved some stuff from `entity-utils.ts` into `entity-column-utils.ts` since things were getting pretty crowded in `entity-utils.ts`.

I also removed the deprecated roster page that wasn't using a view, since I kept having to fix issues in it and it was a waste of time. Nobody should be working out of that page anymore, so it should be fine.


packages/server/src/util/entity-utils.ts
```diff
@@ -389,3 +274,23 @@ export class EntityController<T = any> {
...
 
 }
+
+function formatColumnValue(val: ColumnValue, column: ColumnInfo) {
+  if (val === null) {
+    return val;
+  }
+  switch (column.type) {
+    case ColumnType.String:
+      return val.toString().toLowerCase();
```
jdgarrett, line 284: What is the reasoning behind doing `toLowerCase` here?  It seems that a query such as `column1 = 'AbC'` would fail because it would turn into `column1 = 'abc'`?

packages/shared/src/utils/entity-column-utils.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+}
+
+export function columnTypeToDataType(columnType: ColumnType) {
+  // Get dates and enums as strings.
+  switch (columnType) {
+    case ColumnType.Date:
+    case ColumnType.DateTime:
+    case ColumnType.Enum:
+      return 'string';
```
jdgarrett, line 140: Should this be `ColumnType.String`?

jdgarrett (changes requested):
Overall this looks good to me, I just noticed a couple of small things that potentially need to be looked at.

packages/server/src/api/roster/roster.model.ts
```diff
@@ -55,7 +58,7 @@ export class Roster extends RosterEntity {
     custom: false,
     required: false,
     updatable: false,
-    sql: `COUNT(json_array_length("observation"."custom_columns"->'Details'->'Symptoms') > 0)`,
+    sql: `COUNT(json_array_length("observation"."custom_columns"->'Symptoms') > 0)`,
```
richard3d, line 61: Isn't the correct operator `->>` in this case? Not `->`

PR #256, opened 2021-10-06
Ensure a roster entry is only on one unit per organization.

jdgarrett:
Roster is now unique on EDIPI and organization.  This PR also fixes a few bugs for users with limited permissions which caused requests to go to the server for things they did not have permission for.

### DIFF
--- a/packages/server/src/api/roster/roster-entity.ts
+++ b/packages/server/src/api/roster/roster-entity.ts
@@ -44,13 +44,13 @@ export abstract class RosterEntity extends BaseEntity {
     nullable: false,
     onDelete: 'CASCADE',
   })
-  org!: Org;
+  org?: Org;
 
   @ManyToOne(() => Unit, unit => unit.id, {
     nullable: false,
     onDelete: 'RESTRICT',
   })
-  unit!: Unit;
+  unit?: Unit;
 
   @Index()
   @Column({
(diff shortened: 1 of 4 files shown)


packages/client/src/components/pages/home-page/home-page.tsx
```diff
@@ -139,14 +139,16 @@ export const HomePage = () => {
         await dispatch(Workspace.fetchDashboards(orgId, workspace.id));
       }
 
-      try {
-        await dispatch(OrphanedRecordActions.fetchCount({ orgId: orgId! }));
-      } catch (error) {
-        void dispatch(Modal.alert('Get Orphaned Records', formatErrorMessage(error, 'Failed to get orphaned records')));
+      if (user.activeRole?.role.canManageRoster) {
```
larskendall, line 142: You can potentially make some of these permissions checks a little cleaner by getting the permissions the same way the View component does, like this...

```typescript
  const { canManageGroup } = useAppSelector(UserSelector.role)!;
```

packages/server/src/api/roster/roster-entity.ts
```diff
@@ -41,6 +40,12 @@ export abstract class RosterEntity extends BaseEntity {
   @PrimaryGeneratedColumn()
   id!: number;
 
+  @ManyToOne(() => Org, org => org.id, {
+    nullable: false,
+    onDelete: 'CASCADE',
+  })
+  org!: Org;
```
larskendall, line 47: Since this is a relation without `eager: true` it's best to type it as possibly undefined in case roster entries are queried without it.

```suggestion
  org?: Org;
```

packages/client/src/components/pages/home-page/home-page.tsx
jdgarrett: I am going to leave this one for now, since all the other checks in the file do not follow that syntax yet.

PR #258, opened 2021-10-11
Client actions revamp.

larskendall:
All of the client-side code for handling actions should now be immutable, with actions registered in redux via `entity-action-registry.slice.ts`.

I introduced the concept of a "baked" action, meaning one that's not definable in a data-driven way because it does something custom/complex.

This PR also includes code that I moved/adapted from [this task](https://app.asana.com/0/1188940305683068/1200982891588094) that Martin had in review. Once this stuff gets merged we can just close out the other PR for that task.

I left a couple test actions in, which we should make a task to remove. I figure it can be done as a separate task, so that this PR is still easily testable. I'll make a ticket for it.


packages/client/src/entity-actions/entity-action-baked.ts
```diff
@@ -0,0 +1,85 @@
...
+export const entityActionBakedTableItems: {
+  [K in EntityActionBakedTableItemIds]: EntityActionTableItem;
+} = {
+  actionRosterAddEntry: {
+    type: 'table-button-item-baked',
+    id: 'actionRosterAddEntry',
+    entityType: 'roster',
+    displayName: 'Add Roster Entry',
+    operation: undefined,
+    requiredPermissions: ['canManageRoster'],
```
jdgarrett, line 28: Could these permissions be part of an enumeration somewhere to avoid hard-coded strings?

packages/shared/src/role.types.ts
```diff
@@ -14,8 +14,19 @@ export type RoleSerialized = {
   canManageGroup: boolean;
   canManageRoster: boolean;
   canViewRoster: boolean;
+  canViewObservation: boolean;
```
jdgarrett, line 17: Is this permission being used to hide the observation page if they do not have it?

PR #259, opened 2021-10-12
Fix issues with unit column in roster and observation pages

jdgarrett:
This PR re-adds the unit column to the roster table and allows filtering on that column.  It also removes one of the extra reported unit fields from the observation table, adds the roster unit field, and allows filtering on the roster unit.


larskendall (approved):
Looks great!

PR #260, opened 2021-10-13
Don't show action columns in query builder.

larskendall:
https://app.asana.com/0/1188940305683068/1201140644034607

The filter query builder should no longer show any action columns.


packages/client/src/components/view/view.tsx
```diff
@@ -123,28 +123,30 @@ export default function View(props: ViewProps) {
...
-      }
-
-      return {
-        ...column,
-        type,
-        enumItems,
-      };
-    });
+    return columns
+      .filter(column => !column.action)
```
larskendall, line 127: This is the only new line here. I just reformatted the rest.

PR #261, opened 2021-10-13
Remove Spaces/ES/Kibana.

larskendall:
https://app.asana.com/0/1188940305683068/1200861472713991

A lot of these lines removed are from clearing out the kibana saved object mocks in migrations. A lot of the migrations relying on the WorkspaceTemplate model were broken after clearing this stuff out, and it didn't seem worth the time to actually fix them, so I cleared those migrations out as well (leaving the files for consistency with the database).

Otherwise, I removed anything having to do with Elasticsearch, Kibana, or Spaces. Tests seem to be passing and the app is functioning properly as far as I can tell.


jdgarrett (approved):
Looks good to me!

PR #266, opened 2021-10-15
Fix issues with exporting, join columns, and permissions

jdgarrett:
- Fix issue with join columns appearing in role dialogs
- Fix issue with filtering on aggregated join columns
- Remove test actions
- Fix missing view observation permissions
- Fix data export for roster and observations


mbrinkg8tr (approved):
Changes look good, passes functional testing, unit testing, and code has been eslinted.

PR #267, opened 2021-10-15
Fix issues with table sorting

jdgarrett:
Prevent sorting of action columns and fix issue where sort arrow was not displayed.


mbrinkg8tr (approved):
Passes functional and unit testing, eslint clean

